PR #8, opened 2022-02-11
Added a json body parser size limit as a config option

skid:
This PR adds a config option to the google cloud tasks plugin. The config option controls the body size for the JSON body parser. I'm quite sure that cloud tasks can handle larger payloads than the default size limit setting.


martijnvdbrug:
Some additional info for future reference: 
Max Cloud Tasks size is 1mb: https://cloud.google.com/tasks/docs/quotas
Default BodyParse size is 100kb: https://github.com/expressjs/body-parser#:~:text=Defaults%20to%20true%20.-,limit,Defaults%20to%20'100kb'%20.

Thanks for your contribution :)

PR #45, opened 2022-08-08
invoice-plugin - s3 storage naming and content type

Conor-Burns:
Suggesting simpler naming inside bucket and setting the content type for direct display from s3 url


martijnvdbrug:
I added the tmpfile name to prevent duplicate files/collisions, but I don't think there is a scenario where 2 invoices with the same number can be generated. 

Could you also change the path in the GoogleStorageInvoiceStrategy and the LocalFileStrategy maybe? This way we keep the file paths in sync across different strategies

Conor-Burns:
@martijnvdbrug sure thing :+1: - please review if this is to your linking

martijnvdbrug (approved):

martijnvdbrug:
:ok_hand: published as 3.9.1

PR #81, opened 2023-01-15
Typo Change

iamrahultanwar:
Just a type change in the code


martijnvdbrug:
Thank you for making the effort to contribute :partying_face: It will be published in the next patch or minor version. (Should be somewhere next week)

PR #82, opened 2023-01-15
Feat/stripe scubscription

is0utfitters:
Updated the price calculations so that the Vendure Variant Price of a subscription is always the recurringPrice unless the default downpayment is changed on the orderLine.  This should make it easier for store owners since the Variant Price for a subscription will match as the recurringPrice.


martijnvdbrug (approved):
No tests yet, so living on the edge: merged and will publish it now

PR #84, opened 2023-01-22
feat(shipping): variant weight overrides

kyunal:
This change allows users to change weight of a product on a variant basis. If no variant weight is specified, the product weight is used. If this still doesn't exist then 0 is used (like before).

This probably doesn't need its own release, as I am intending to add more physical property features in the near future.


martijnvdbrug (approved):
Perfect! Additional testcases and everything. Thanks for your contribution

martijnvdbrug:
Merged. I will publish it in the next version with other changes, unless you need it.

martijnvdbrug:
@kyunal This is published under 4.9.0

PR #160, opened 2023-05-16
Feat(Webhook): Support multiple outgoing webhooks

martijnvdbrug:
# Description

The Webhook Plugin has been extended to support multiple webhooks per channel and per event. An administrator can now create multiple outgoing webhooks for multiple events via the Admin UI.

# Breaking changes

The DB needs to be migrated, because a new entity `Webhook` has been created. The old entity will be deleted from the DB.
If you already use this plugin, you need to manually recreate the hooks:

* Check what URL is triggered for what event, and note it down somewhere
* Install the plugin, start the server, and go to `Settings > Webhook` in the Admin UI
* Create the hook. You can leave the `Transformer` field blank: the plugin will send an empty post without a transfomer.

# Screenshots

image

# Checklist

Please make sure you've done the following checks:

- [x] I have set a clear title
- [x] My PR is small and contains only a single feature. (Split into multiple PR's if needed)
- [x] I have added or updated test cases for important functionality
- [x] I have updated the README if needed
- [x] I have reviewed my own PR
- [x] I have fixed all typo's and have removed unused or commented code

PR #167, opened 2023-05-24
fix: webhook plugin - circular dependency

seminarian:
# Description
In certain cases when working with custom populate scripts for multi-channel seeds I ran in this obscure error whenever the I tried to get the webhookService in that script. Turns out the WebhookService was undefined because of a circular dependency.
Also a circular dependency regarding the Permission acted up.

Hope you can approve this change! :) It shouldn't change anything functionally!

Options for the plugin are now injected via NestJs service container instead of importing the plugin. And the Permission is now defined in he resolver.
# Breaking changes

Does this PR include any breaking changes we should be aware of?
No

### DIFF
--- a/packages/vendure-plugin-webhook/src/api/webhook.resolver.ts
+++ b/packages/vendure-plugin-webhook/src/api/webhook.resolver.ts
@@ -1,7 +1,12 @@
 import { Args, Mutation, Query, Resolver } from '@nestjs/graphql';
 import { Allow, Ctx, RequestContext } from '@vendure/core';
 import { WebhookService } from './webhook.service';
-import { webhookPermission } from '../index';
+import { PermissionDefinition } from '@vendure/core';
+
+export const webhookPermission = new PermissionDefinition({
+  name: 'SetWebhook',
+  description: 'Allows setting a webhook URL',
+});
 
 /**
  * Graphql resolvers for retrieving and updating webhook for channel
--- a/packages/vendure-plugin-webhook/src/api/webhook.service.ts
+++ b/packages/vendure-plugin-webhook/src/api/webhook.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, OnApplicationBootstrap } from '@nestjs/common';
+import { Injectable, Inject, OnApplicationBootstrap } from '@nestjs/common';
 import { ModuleRef } from '@nestjs/core';
 import {
   EventBus,
@@ -8,7 +8,8 @@ import {
   VendureEvent,
 } from '@vendure/core';
 import { WebhookPerChannelEntity } from './webhook-per-channel.entity';
-import { WebhookPlugin } from '../webhook.plugin';
+import { PLUGIN_INIT_OPTIONS, loggerCtx } from '../constants';
+import { WebhookPluginOptions } from './webhook-plugin-options';
 import fetch from 'node-fetch';
 import { loggerCtx } from '../constants';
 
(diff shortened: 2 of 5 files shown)


martijnvdbrug:
@seminarian Thanks! Is this urgent for you? I am also in the middle of an upgrade of the webhook plugin, so I prefer to merge that first (it has the same fixes ;-) ).

Check it out here: https://github.com/Pinelab-studio/pinelab-vendure-plugins/pull/160
There are some breaking changes (new entities). Upgrade path is now to manually recreate webhooks, does that work for you, or do we need some script?

seminarian:
@martijnvdbrug Great to hear Martijn!

There's no rush, ideally when the plugins are ported to v2 these would be included then I can delete my local copy of the plugin.
I don't think manual migration will be an issue.

Got a question though:
- Will it still be easy to configure a webhook for all the Events (which are now defined via the plugin options) or would there be a need to manually list them all up for every webhook that needs them?

martijnvdbrug:
That's a good point, I actually also need that. Maybe we can add a wildcard * to trigger on all events

martijnvdbrug:
Better would be to be able to select multiple events

seminarian:
> Better would be to be able to select multiple events

Select multiple events and a just "Select All" option is the perfect merge.
Will you allow users to select events that are not defined in the plugin options or not? That could be a plugin option flag as well?
How do you see this matter?

About those transformers, am I correct to assume that those that you can fill in there must be pre-configured in the plugin configuration?

martijnvdbrug:
Yep, both transformers and events need to be configured in the plugins init() before an admin can select them. This is because we need to know what events to listen for on application bootstrap

seminarian:
> Yep, both transformers and events need to be configured in the plugins init() before an admin can select them. This is because we need to know what events to listen for on application bootstrap

Oh right, makes sense!

martijnvdbrug:
@seminarian So, we've finished the new plugin, it's merged in [this PR](https://github.com/Pinelab-studio/pinelab-vendure-plugins/pull/160)

We've decided that we won't allow multiple events per webhook for now, because it adds a lot of complexity (also UI wise). We looked at Zappier, and they also keep it simple and readable by just having 1 event attached to 1 webhook. 

We are working on a `duplicate hook` button, which makes it easier to create a hook for multiple events. So, if you want the same hook/url for multiple events, you'd just create multiple hooks via the UI. The plus side of this is that you can still choose different transformers per event.

The plugin is now published as 7.0.0-v2.1, and supports Vendure V2. Do you need this available for V1.X?

martijnvdbrug:
> @martijnvdbrug Thank you for the update! We are already on V2 so we don't need the v1 version :)
> 
> If I understood it correctly, we currently use multiple channels and listen to 21 Evenst for 1 webhook: we will need to create 21 webhooks per channel with the new plugin version?

The answer now is: yes, unfortunately that's the only option. I'm open to suggestions and PR's on this plugin though!

seminarian:
> > @martijnvdbrug Thank you for the update! We are already on V2 so we don't need the v1 version :)
> > If I understood it correctly, we currently use multiple channels and listen to 21 Evenst for 1 webhook: we will need to create 21 webhooks per channel with the new plugin version?
> 
> @seminarian The answer now is: yes, unfortunately that's the only option. I'm open to suggestions and PR's on this plugin though!

Alright thanks for your efforts and clarifications! I do think we can probably work around it or even create a small script so not such a big deal!

martijnvdbrug:
> > > @martijnvdbrug Thank you for the update! We are already on V2 so we don't need the v1 version :)
> > > If I understood it correctly, we currently use multiple channels and listen to 21 Evenst for 1 webhook: we will need to create 21 webhooks per channel with the new plugin version?
> > 
> > 
> > @seminarian The answer now is: yes, unfortunately that's the only option. I'm open to suggestions and PR's on this plugin though!
> 
> Alright thanks for your efforts and clarifications! I do think we can probably work around it or even create a small script so not such a big deal!

Allright, closing this one! Let me know if you run into any isssues.

PR #188, opened 2023-06-29
Cloud tasks log failed jobs

dalyathan:
# Description

[`CloudTasksJobQueueStrategy`](https://github.com/Pinelab-studio/pinelab-vendure-plugins/blob/master/packages/vendure-plugin-google-cloud-tasks/src/cloud-tasks-job-queue.strategy.ts) now implements [`InspectableJobQueueStrategy`](https://docs.vendure.io/typescript-api/job-queue/inspectable-job-queue-strategy/), making it  possible  to do [`JobRecord`](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/plugin/default-job-queue-plugin/job-record.entity.ts) related queries when it is the designated [`jobQueueStrategy`](https://docs.vendure.io/typescript-api/job-queue/job-queue-options/#jobqueuestrategy). So when a `CloudTaskMessage` succeeds or fails it will be recorded as a [`JobRecord`](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/plugin/default-job-queue-plugin/job-record.entity.ts). Also each [`JobRecord`](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/plugin/default-job-queue-plugin/job-record.entity.ts) related queries inside [`JobResolver`](https://github.com/vendure-ecommerce/vendure/blob/master/packages/core/src/api/resolvers/admin/job.resolver.ts) now resolve to their corresponding queries inside of [`CloudTasksJobQueueStrategy`](https://github.com/Pinelab-studio/pinelab-vendure-plugins/blob/master/packages/vendure-plugin-google-cloud-tasks/src/cloud-tasks-job-queue.strategy.ts).
# Breaking changes

This PR doesn't include any breaking changes.

# Screenshots

Couldn't add screenshots since `yarn start` requires working Google Cloud credentials.

# Checklist

Please make sure you've done the following checks:

- [X] I have reviewed my own PR
- [X] I have fixed all typo's and have removed unused or commented code


packages/vendure-plugin-google-cloud-tasks/src/cloud-tasks-job-queue.strategy.ts
```diff
@@ -53,7 +161,7 @@ export class CloudTasksJobQueueStrategy implements JobQueueStrategy {
           `Added job with retries=${cloudTaskMessage.maxRetries} to queue ${queueName}: ${cloudTaskMessage.id} for ${task.httpRequest.url}`,
           CloudTasksPlugin.loggerCtx
         );
-        return new Job({
+        const successfullJob = new Job({
```
martijnvdbrug, line 164: (I will pick up these changes, because I also have to test it locally anyway. But for future reference:)

I think this isn't a successfull job yet, it only pushed a job successfully to the Queue. We actually want to log it when the handling of the job succeeded, which happens in the controller.

packages/vendure-plugin-google-cloud-tasks/src/cloud-tasks-job-queue.strategy.ts
```diff
@@ -1,21 +1,125 @@
...
     this.client = new CloudTasksClient();
   }
 
+  async findOne(id: ID): Promise<Job<any> | undefined> {
+    if (!this.connection) {
+      throw new UserInputError('TransactionalConnection is not available');
+    }
+    const jobRecord = await this.connection
+      .getRepository(JobRecord)
+      .findOne({ where: { id } });
```
martijnvdbrug, line 46: These are deprecated. because we dont have a `ctx` available, we can use this:
```suggestion
    const jobRecord = await this.connection.rawConnection
      .getRepository(JobRecord)
      .findOne({ where: { id } });
```

martijnvdbrug (changes requested):
Looking good already. I still have to finalize some things, will merge this one monday.

image

martijnvdbrug (approved):
Looks good!

Issue #244, opened 2023-08-16
Also return subscriptionPricing in the admin API for the Stripe Subscription PLugin


dalyathan:
**Is your feature request related to a problem? Please describe.**
In the Shop-api we return the following data on an order line:
```
 extend type OrderLine {
    subscriptionPricing: StripeSubscriptionPricing
  }
```

**Describe the solution you'd like**
We would also like to return this data in the admin graphql api.

PR #245, opened 2023-08-16
Return subscriptionPricing also on admin api

dalyathan:
# Description

As requested on #244, this changes should add  `subscriptionPricing` to `OrderLine` by extending the `OrderLine` gql type on admin api.

# Breaking changes

This PR wont introduce any breaking changes.

# Checklist

:pushpin: Always:

:zap: Most of the time:

- [ ] Have you correctly increased the version number in `package.json` to the next [patch/minor/major](https://semver.org/#summary)?
- [ ] Have you added your changes to the [changelog](https://keepachangelog.com/en/1.0.0/)?

### DIFF
--- a/packages/vendure-plugin-stripe-subscription/src/api/graphql-schemas.ts
+++ b/packages/vendure-plugin-stripe-subscription/src/api/graphql-schemas.ts
@@ -48,15 +48,13 @@ const sharedTypes = gql`
     useProration: Boolean
     autoRenew: Boolean
   }
+  extend type OrderLine {
+    subscriptionPricing: StripeSubscriptionPricing
+  }
 `;
 
 export const shopSchemaExtensions = gql`
   ${sharedTypes}
-
-  extend type OrderLine {
-    subscriptionPricing: StripeSubscriptionPricing
-  }
-
   type StripeSubscriptionPricing {
     variantId: String!
     pricesIncludeTax: Boolean!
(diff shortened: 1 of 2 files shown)


martijnvdbrug (changes requested):
That was easy :-) 

Minor changes requested:

1. Could you add a very simple test that checks if the `subscriptionPricing` object is indeed available via the admin API?

2. Can you change the version number? I think this is a minor update

3. Can you rename `ShopOrderLinePricingResolver` to something more suitable? It's not only a shop-api resolver anymore :-)

packages/vendure-plugin-stripe-subscription/src/api/stripe-subscription.controller.ts
```diff
@@ -115,14 +119,22 @@ export class ShopResolver {
...
   @ResolveField()
   async subscriptionPricing(
     @Ctx() ctx: RequestContext,
     @Parent() orderLine: OrderLineWithSubscriptionFields
   ): Promise<StripeSubscriptionPricing | undefined> {
+    if (ctx.apiType !== 'shop') {
+      await this.entityHydrator.hydrate(ctx, orderLine, {
+        relations: ['productVariant'],
+      });
+    }
```
martijnvdbrug, line 137: Ahh, this relation is not available in the admin-api? You can still do the hydration without if statement, because the EntityHydrator will check if the relation is already available when you call `hydrate`
```suggestion
    await this.entityHydrator.hydrate(ctx, orderLine, {
      relations: ['productVariant'],
    });
```

martijnvdbrug (approved):
Nice work

PR #249, opened 2023-08-21
fix: send invoice data to picqer if billing address is set

nicoes:
# Description

Fix for sending invoice data if billing address is set. Do not use shipping address instead.

### DIFF
--- a/packages/vendure-plugin-picqer/src/api/picqer.service.ts
+++ b/packages/vendure-plugin-picqer/src/api/picqer.service.ts
@@ -1073,10 +1073,25 @@ export class PicqerService implements OnApplicationBootstrap {
     customerId?: number
   ): OrderInput {
     const shippingAddress = order.shippingAddress;
-    // Check billing address existance based on PostalCode
-    const billingAddress = order.billingAddress.postalCode
-      ? order.billingAddress
-      : order.shippingAddress;
+
+    let invoiceData: Partial<OrderInput> = {
+      invoicename:
+        order.billingAddress?.company ||
+        order.billingAddress?.fullName ||
+        undefined,
+      invoicecontactname: order.billingAddress?.fullName || undefined,
+      invoiceaddress:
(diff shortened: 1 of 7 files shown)


packages/vendure-plugin-picqer/src/api/picqer.service.ts
```diff
@@ -1073,10 +1073,25 @@ export class PicqerService implements OnApplicationBootstrap {
...
+      invoicename:
+        order.billingAddress?.company ||
+        order.billingAddress?.fullName ||
+        undefined,
+      invoicecontactname: order.billingAddress?.fullName || undefined,
+      invoiceaddress:
+        order.billingAddress?.streetLine1 || order.billingAddress?.streetLine2
+          ? [order.billingAddress.streetLine1, order.billingAddress.streetLine2]
+              .join(' ')
+              .trim()
```
nicoes, line 1087: Join + trim because streetline 2 can be optional

packages/vendure-plugin-picqer/test/picqer.spec.ts
```diff
@@ -215,6 +228,13 @@ describe('Picqer plugin', function () {
     expect(picqerOrderRequest.deliveryzipcode).toBeDefined();
     expect(picqerOrderRequest.deliverycity).toBeDefined();
     expect(picqerOrderRequest.deliverycountry).toBe('NL');
+    expect(picqerOrderRequest.invoicename).toBe("Martinho's friend");
+    expect(picqerOrderRequest.invoicecontactname).toBe("Martinho's friend");
+    expect(picqerOrderRequest.invoicecountry).toBe('NL');
+    expect(picqerOrderRequest.invoiceaddress).toBe('Remote location 123');
+    expect(picqerOrderRequest.invoicezipcode).toBe('1111AB');
+    expect(picqerOrderRequest.invoicecity).toBe('Faraway');
+    expect(picqerOrderRequest.invoicecity).toBe('Faraway');
```
martijnvdbrug, line 237: Duplicate
```suggestion
```

PR #256, opened 2023-09-13
Order Client Loading States

martijnvdbrug:
# Description

This PR aims to implement global loading and error states for the active order store and the current user store.
Vue example:
```vue
<template>
  <div>
    <h1 v-if="loading">Loading...</h1>
    <h1 v-if="error">Something went wrong!</h1>
    <h1 v-if="data">Active order: {{ data.code }}</h1>
  </div>
</template>
<script setup lang="ts">
  import { VendureOrderClient } from "vendure-order-client";
  import { useStore } from "@nanostores/vue";

  const client = new VendureOrderClient(
    "http://localhost:3050/shop-api",
    "your-channel-token"
  );

  const { data, loading, error } = useStore(client.$activeOrder);
  await client.getActiveOrder();
</script>
```

:warning: Still TODO:

* For queries and mutations that handle activeOrders (like `addItemToOrder`, `applyCouponCode`, etc), you can use the decorator `@HandleLoadingState('$activeOrder')` to handle loading and error states for `$activeorder`
* For queries and mutations that handle currentUser (like `login`), you can use `@HandleLoadingState('$currentUser')` so that it handles loading and error states for `$currentUser`
* For each function/mutation:
  * Add the decorator `@HandleLoadingState('$currentUser')` or `@HandleLoadingState('$activeOrder')`
  * Use `setResult(this.$activeOrder, activeOrder);` or `setResult(this.$currentUser, currentUser);` to store the result (or ErrorResult) in the store.
  *  Checkout `getActiveOrder()` as example, that should already be correctly implemented (untested)
  * And last, do `this.throwIfErrorResult(activeOrder as ActiveOrder<A>);` to return either an `ActiveOrder` or throw if an ErrorResult was returned.
* All `validateCurrentUser` and `validateOrder` can be replaced with `throwIfErrorResult`
* Testing: `@nanostores/vue` needs to be installed, so that we can test if loading and result states are correctly set. I guess we can use `const { data, loading, error } = useStore(orderClient.$activeOrder);` in our tests.
  * Testing the error handling once, for example in by adding a non-existent item to order is good enough, no need to test error state for each function, because it might be hard to mock errors in some cases (like for `getActiveOrder()`)
  * Each test should test for the correct `loading` and `data` result though

:warning: All of this is untested, so please let me know if you run into issues or if things aren't clear.

# Breaking changes

* Consumers can now destructure the store to get data, loading and error states:
```diff
import { useStore } from '@nanostores/vue'

- const activeOrder = useStore(orderClient.activeOrder);
+ const { data, loading, error } = useStore(orderClient.$activeOrder);
```

# Checklist

:pushpin: Always:

:zap: Most of the time:

- [X] Have you correctly increased the version number in `package.json` to the next [patch/minor/major](https://semver.org/#summary)?
- [X] Have you added your changes to the changelog? [See this example](https://github.com/Pinelab-studio/pinelab-vendure-plugins/blob/main/packages/vendure-plugin-invoices/CHANGELOG.md)


martijnvdbrug:
Nice work @dalyathan, thanks

packages/vendure-order-client/test/vendure-order-client.spec.ts
```diff
@@ -135,10 +138,31 @@ describe('Vendure order client', () => {
...
+      let wasSetToTrue = false;
+      let wasFinallySetToFalse = false;
+      listenKeys(client.$activeOrder, ['loading'], (currentActiveOrderData) => {
+        if (currentActiveOrderData.loading) {
+          wasSetToTrue = true;
+        } else {
+          if (wasSetToTrue) {
+            wasFinallySetToFalse = true;
+          }
+        }
```
martijnvdbrug, line 150: These else statements could be simplified into:

```suggestion
        } else if (wasSetToTrue) {
            wasFinallySetToFalse = true;
        }
```

BUT, another option would be to not use `listenKeys` at all:

```ts
      expect(client.$activeOrder.value?.loading).toBe(false);
      const promise = client.removeOrderLine('T_1');
      expect(client.$activeOrder.value?.loading).toBe(true);
      await promise;
      expect(client.$activeOrder.value?.loading).toBe(false);
```
It does require a bit of work to refactor, but I do find the code a bit more followable. I leave it up to you, both approaches are ok, but if you choose the `listenKeys` options we need to simplify the if/else statement like above.

I do like the custom `expect` messages, I was actually looking for something like this the other day, didnt know this worked :open_hands:

packages/vendure-order-client/test/vendure-order-client.spec.ts
```diff
@@ -71,13 +72,37 @@ describe('Vendure order client', () => {
...
+  }
+
+  async function testLoadingState(
+    awaitAbleFunction: SearchFn,
+    store: MapStore
+  ): Promise<any> {
+    expect(store.value?.loading).toBe(false);
+    const promise = awaitAbleFunction();
+    expect(store.value?.loading).toBe(true);
+    return await promise;
```
martijnvdbrug, line 97:
```suggestion
   const result = await promise;
   expect(store.value?.loading).toBe(false);
   return result;
```

PR #262, opened 2023-09-25
feat: allow option to trigger additional actions when a processing er…

nicoes:
# Description

We need to inspect the error that was thrown while processing a job. This error has specific details that I want to log. Since the cloud tasks plugin catches all errors this is hard to do now. If I can pass a hook that will be called with the original error then I can log / inspect that.

I haven't added test yet or a package version, because I wanted to check the rough idea with you first


martijnvdbrug:
Great addition thanks :-) 

I've made some minor changes to your PR:

* Moved handler execution to execute before `res.send()`
* Renamed `onProcessingError` to `onJobFailure`
* Allow return of promises from onJobFailure + wrapped in try/catch, because we shouldn't allow a custom onJobFailure to break our flow
* Updated package.json + changelog

martijnvdbrug (approved):
Thanks!

PR #272, opened 2023-10-20
Fix Popularity Plugin Error on Postgres Databases

dalyathan:
# Description

When calculating product popularity scores, we were querying unused columns. It worked fine on `mysql` and `sqlite` databases but failed on `postgres`. This PR removes those select statements.
# Breaking changes

Does this PR include any breaking changes we should be aware of? **No**

# Screenshots

Add screenshots if needed.

# Checklist

:pushpin: Always:

:zap: Most of the time:

- [X] Have you correctly increased the version number in `package.json` to the next [patch/minor/major](https://semver.org/#summary)?
- [X] Have you added your changes to the changelog? [See this example](https://github.com/Pinelab-studio/pinelab-vendure-plugins/blob/main/packages/vendure-plugin-invoices/CHANGELOG.md)


martijnvdbrug (changes requested):
Good catch! But, can you revert the yarn.lock? It seems there are some postgress dependencies in there that are not in the package.json (and probably should not be in there)

martijnvdbrug (approved):

martijnvdbrug (approved):
Nice catch, how did you find this issue?

I've made a small change: https://github.com/Pinelab-studio/pinelab-vendure-plugins/pull/272/commits/274a58b30278435247c75632651095bab384955e#diff-96b08916769af8a4454a284a4d800ca82ac91d1ac226ea38abb867b7e54732ab

Since `variantsPartialInfo2` is not really a good name. I've renamed it to `variantsPartialInfoResults` as it's the result of the `variantsPartialInfo` query.

Issue #316, opened 2023-12-29
Anonymized order plugin exposes customer details


martijnvdbrug:
**Which plugin/repository is the issue about?**
[anonymized-order]

**Describe the bug**
We filter out an orders sensitive information like `shippingAddress` and `customer`, but with a nested query we can, for example get all orders of a customer.

**To Reproduce**
Query an anonymized order with the following query:

```graphql
{
  anonymizedOrder(orderCode:"CG-201023-UBN8K", emailAddress:"martijn@pinelab.studio") {
    id
    createdAt
    updatedAt
    lines {
      id
      order {
        customer {
          firstName
          lastName
          orders {
            items {
              id
              code
            }
          }
        }
      }
    }
  }
}
```
This exposes all orders of a customer, including all sensitive details!

**Expected behavior**
The sensitive fields should not be exposed. We should also nullify `order.lines.order`. Are there any other fields that can cause this kind of unwanted data exposure?

Extra request: Can we nullify shippingAddress instead of returning an empty object? It's an optional field on an order, so we can just set it to undefined or null

dalyathan:
yes, nullifying order.lines.order will do the trick. I will look to see if there are any more fields that could expose sensitive data.
I don't think we can nullify shippingAddress as it is a non-nullable field for Order. I tried and ts complained

PR #317, opened 2023-12-29
Fix exposed fields bug

dalyathan:
# Description

The changes in this PR fix the bug reported in [here](https://github.com/Pinelab-studio/pinelab-vendure-plugins/issues/316)
# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


packages/test/src/generated/shop-graphql.ts
```diff
@@ -3424,6 +3424,7 @@ export const OrderFields = gql`
       type
     }
   }
+  couponCodes
```
martijnvdbrug, line 3427: Please make sure you edit the .graphql files and not the generated files liek this one. This will get overwritten when types are regenerateed.

Te correct flow is:

1. Add fields in `test/src/shop.graphql`
2. Start the dev server with `yarn start` inside `/packages/test`folder
3. In another terminal window, run `yarn generate`, also in the `/test` folder.

martijnvdbrug (approved):
Nice work. I've pushed some minor change, so this is just FYIfor other pr's:

1. I've made this a patch instead of a minor bump (1.0.1 insteead of 1.1.0) because its a bugfix
2. Moved the new version in changelog to the top: Newest version on top is more readable for our consumers I think
3. Please make sure you edit the `.graphql` files and not the generated files in test package. This will get overwritten when types are regenerated.

Te correct flow is:

- Add fields in `test/src/shop.graphql`
- Start the dev server with yarn start` inside `/packages/test`
- In another terminal window, run `yarn generate`, also in the /test folder.

packages/test/src/generated/shop-graphql.ts
dalyathan: okay

PR #325, opened 2024-01-10
Pinelab plugin admin components

dalyathan:
# Description

The changes in this PR create a new plugin called `pinelab-plugin-admin-components`.
# Breaking changes

# Screenshots

Screenshot 2024-01-10 at 3 46 24 AM

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


package.json
```diff
@@ -107,7 +107,8 @@
           "stripe-subscription",
           "variant-bulk-update",
           "vendure-order-client",
-          "webhook"
+          "webhook",
+          "pinelab-plugin-admin-components"
```
martijnvdbrug, line 111:
```suggestion
          "admin-components"
```

packages/pinelab-plugin-admin-components/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# 1.0.0 (2024-01-10)
+
+- Creatd the plugin
```
martijnvdbrug, line 3:
```suggestion
- Abstracted HTML editor component in to admin components package
```

martijnvdbrug (changes requested):
I was expecting to only package Angular components for reuse in the invoices plugin and the picklist plugin, something like this:
```
import { HTMLTemplateViewer } from '@pinelab/admin-components';

// in you r Angular (admin UI) html
<pinelab-html-editor />
`;

This PR duplicates the backend and everything as well, which is not what we want.

I suggest we just duplicate the html part in the picklist plugins and invoices plugin for now. What do you think?

(Sorry for the unclarity and the work you've already put in)

PR #327, opened 2024-01-12
EligibleGift Query Bug Fix

### LINKED ISSUE #326: Selectable gifts plugin returns variants without stock
The plugin returns variants as selectable gifts via the query `eligibleGifts`. This query now also returns variants that are out of stock. It should only return variants that are in stock!

Docs can be found here: https://pinelab-plugins.com/plugin/vendure-plugin-selectable-gifts/

dalyathan:
# Description

The changes in this PR fix the bug reported in #326 . Closes #326 

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


martijnvdbrug (changes requested):
Some clarification requested :-)

packages/vendure-plugin-selectable-gifts/src/api/gift.service.ts
```diff
@@ -47,7 +47,20 @@ export class GiftService {
     if (!variantIds.length) {
       return [];
     }
-    return await this.variantService.findByIds(ctx, variantIds);
+    return this.connection
```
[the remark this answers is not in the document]
dalyathan: yeah your right, those fields will most likely be empty when querying with the raw collection. I was thinking of using `TranslatorService` and `ProductPriceApplicator` to populate those fields, and also adding some checks. What do you think?

packages/vendure-plugin-selectable-gifts/src/api/gift.service.ts
```diff
@@ -47,7 +50,30 @@ export class GiftService {
...
+      .andWhere('variant.deletedAt IS NULL')
+      .andWhere('channel.id = :channelId', { channelId: ctx.channelId })
+      .getMany();
+    const translatedVariants = variants.map((v) =>
+      this.translatorService.translate(v, ctx)
+    );
+    const translatedVariantsWithPrice = translatedVariants.map((v) =>
+      this.priceApplicator.applyChannelPriceAndTax(v, ctx)
+    );
+    return Promise.all(translatedVariantsWithPrice);
```
martijnvdbrug, line 76: I think awaiting both promises this way could cause race conditions:
(Simplified example)
1. Example A is `{id: 1}`
2. Variant translation is started, but not awaited (so in progress)
3. Simultaneously, the applyPrices is being started, also not awaited
4. Translation is finished, so it will set the variant object now to `{ id: 1, name: 'Some translated name'}`
5. ApplyPrices is now also finished, but it sets the object to `{ id: 1, price: 1234}`. Notice that the translated name is now gone, because the input of applyPrices was the variant without translated fields.

We can't be 100% sure this happens without knowing the functions internals, but its best to `Promise.all` all translations first before applying prices.

martijnvdbrug (changes requested):
Another minor change requested

packages/vendure-plugin-selectable-gifts/src/api/gift.service.ts
dalyathan: I don't think `translatorService.translate` returns a promise. It uses [translateDeep](https://github.com/vendure-ecommerce/vendure/blob/8a3b5f818046c0ae83099781d3763f17783f376e/packages/core/src/service/helpers/utils/translate-entity.ts#L96) and I think it returns a `Translated<T>`

martijnvdbrug: I misread the last line as `return Promise.all(translatedVariantsWithPrice, translatedVariants);`!

You are totally right, my bad. Well so far for my story about race conditions :laughing:

martijnvdbrug (approved):
YES!

PR #331, opened 2024-01-15
feat(vendure-order-client): updated the getOrderByCode for polling or…

ZackNew:
- updated the getOrdeByCode to poll for 30 seconds and return the order if state is paymentSettled or PaymentAuthorized

### DIFF
--- a/packages/vendure-order-client/CHANGELOG.md
+++ b/packages/vendure-order-client/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.8.1 (2023-12-01)
+
+- Update the getOrderByCode to be polling for 30 seconds
+
 # 2.8.0 (2024-01-04)
 
 - Added `getActiveCustomer` and `logout()` functions;
--- a/packages/vendure-order-client/CHANGELOG.md
+++ b/packages/vendure-order-client/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 2.8.1 (2023-12-01)
 
-- Update the getOrderByCode and add pollOrderByCode function
+- Added `pollOrderByCode` to poll for order every second
 
 # 2.8.0 (2024-01-04)
 
--- a/packages/vendure-order-client/CHANGELOG.md
+++ b/packages/vendure-order-client/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 2.8.1 (2023-12-01)
 
-- Update the getOrderByCode to be polling for 30 seconds
+- Update the getOrderByCode and add pollOrderByCode function
 
 # 2.8.0 (2024-01-04)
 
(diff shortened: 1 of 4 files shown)


martijnvdbrug (changes requested):
Sorry for the late response. Good addition, but I would like to request a change:

 Can we create a new method, so that we can also still use the `getOrderByCode` if needed ? I would propose something like `pollOrderByCode(code: string, pollFor? : number = 10)`. The poll method would also handle loading state with `@HandleLoadingState('$activeOrder')`

So we would have both methods `pollOrderByCode` for polling and the raw `getOrderByCode` when a consumer doesn't want to poll for X seconds.

packages/vendure-order-client/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.8.1 (2023-12-01)
+
+- Update the getOrderByCode to be polling for 30 seconds
```
martijnvdbrug, line 3:
```suggestion
- Added `pollOrderByCode` to poll for order every second
```

packages/vendure-order-client/src/vendure-order-client.ts
```diff
@@ -364,9 +364,36 @@ export class VendureOrderClient<A = unknown> {
       this.queries.GET_ORDER_BY_CODE,
       { code }
     );
+    if (
```
martijnvdbrug, line 367: GetOrderByCode should not do this check. it should be a dumb function: just return the result. This logic can be added to `pollOrderByCode`: if not settled or authorized, poll again

packages/vendure-order-client/src/vendure-order-client.ts
```diff
@@ -364,9 +364,36 @@ export class VendureOrderClient<A = unknown> {
...
+    let pollingCount = 0;
+    while (!order) {
+      if (pollingCount > pollFor) {
+        throw Error(`Order not settled after polling ${pollFor} times`);
+      }
+      order = await this.getOrderByCode(code);
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+      pollingCount++;
+    }
+    return order;
```
martijnvdbrug, line 394: Maybe it's a good idea to accept an extra argument:

```ts
async pollOrderByCode(
    code: string,
    pollFor: number = 10,
    acceptedStates: string[] = ['PaymentSettled', 'PaymentAuthorized']
  )
```

This way we can also pass in `AwaitingPrepress` as extra state in the CG project, because thats also a valid state, but only in the CG project.

We would then use the states in this function: 

```ts
if ( !acceptedStates.includes(order.state) {
   // poll again
}

martijnvdbrug (changes requested):
Last changes!

ZackNew:
added argument called acceptedStates to take accepted states and will keep polling while the order state is not included in the passed accepted states.

martijnvdbrug (approved):
Yep!

PR #349, opened 2024-01-29
Vendure Shipping Extensions Plugin

### LINKED ISSUE #347: Distance based shipping calculator
For the CG project, we would like to calculate shipping based on the distance between the store and the customer's address.

We can create a custom shipping calculator:
* The calculator takes the arguments
   * `store latitude`
   * `store longitude`
   * `price per KM`
 * We will calculate the distance between 2 geolocations using [this method](https://stackoverflow.com/questions/36429243/calculating-the-distance-between-two-postcodes-using-javascript), so we dont need an external API for this.
 * This plugin should be extensible, so it should hold a strategy to transform an order's shipping address into lat/lon values, something like `getGeoLocationForAddress(orderAddress): Promise<GeoLocation>` (geolocation=lat/lon)
 * For the CG project, we could then implement the strategy to use [an API](https://www.freemaptools.com/convert-uk-postcode-to-lat-lng.htm) to transform the UK address into a lat/lon value, so that the plugin can calculate the distance. But, for other countries we could implement our own strategy.
 * We should include the UK postalcode-to-geo strategy in this plugin, I am not sure what API to use. Google is your friend :-) Free is good, but if it has to be a paid API then it is what it is. Please choose a widely supported well known API
* This calculator should be included in this plugin: https://github.com/Pinelab-studio/pinelab-vendure-plugins/tree/main/packages/vendure-plugin-shipping-by-weight-and-country
* We should however renamed this plugin to `vendure-plugin-shipping-extensions` to make it more generic. I was thinking this flow for renaming:
  * Copy the plugin to new folder with new name (update package.json name)
  * Publish the new renamed plugin
  * Update the README of the old plugin, with a clickable link to the new plugin on NPM
  * Publish the old plugin with the updated readme
  * Deprecate the old plugin (I can do this manually via CLI)
  * Delete old plugin from the monorepo

dalyathan:
# Description

The changes in this PR implement and close #347 
# Breaking changes

# Screenshots

Screenshot 2024-01-29 at 3 05 55 AM

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


packages/vendure-plugin-shipping-extensions/README.md
```diff
@@ -0,0 +1,64 @@
...
+This plugin does two things in general
+
+- adds a shipping eligibility checker to Vendure that checks the total weight and the shipping country of an
+  order, to verify if a shipping method is eligible for a given order.
+- introduces a distance based (`ShippingCalculator`)[https://docs.vendure.io/reference/typescript-api/shipping/shipping-calculator/], based on a configurable `OrderAddressToGeolocationConversionStrategy` used to convert the `shippingAddress` of an `Order` to geographic latitudinal and longitudinal values.
+
+The weight of a product can be configured on the customfield `Product.weight`. You can configure the units to be in KG,
+grams or whatever unit you like.
+
+A Custom `OrderAddressToGeolocationConversionStrategy` can be configured by as follows:
```
martijnvdbrug, line 14: Maybe a small explanation here? Something in the lines of what this strategy can be used for

packages/vendure-plugin-shipping-extensions/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# 1.0.0 (2024-01-29)
```
martijnvdbrug, line 1: Can you copy the changelog and versioning of the previous plugin here? It's basicly the same plugin, but renamed.

Make it a new major version, so that we can use a patch or minor version to deprecate the old plugin on NPM

packages/vendure-plugin-shipping-extensions/README.md
```diff
@@ -0,0 +1,64 @@
...
+  1100 grams
+
+## Getting started
+
+1. Add the following to the plugins in `vendure-config.ts`:
+
+```ts
+plugins: [
+  ...
+    ShippingByWeightAndCountryPlugin.init({
```
martijnvdbrug, line 39:
```suggestion
    ShippingExtensionsPlugin.init({
```

packages/vendure-plugin-shipping-extensions/src/distance-based-shipping-calculator.ts
```diff
@@ -0,0 +1,71 @@
...
+      ui: { component: 'number-form-input', min: -90, max: 90 },
+      label: [{ languageCode: LanguageCode.en, value: 'Store Latitude' }],
+    },
+    storeLongitude: {
+      type: 'float',
+      ui: { component: 'number-form-input', min: -180, max: 180 },
+      label: [{ languageCode: LanguageCode.en, value: 'Store Longitude' }],
+    },
+    pricePerKm: {
+      type: 'float',
```
martijnvdbrug, line 32: Prices are `int` in Vendure :-) The currency form input handles the floating point I think

packages/vendure-plugin-shipping-extensions/src/distance-based-shipping-calculator.ts
```diff
@@ -0,0 +1,71 @@
...
+    taxRate: {
+      type: 'float',
+      ui: { component: 'number-form-input', suffix: '%', min: 0, max: 100 },
+      label: [{ languageCode: LanguageCode.en, value: 'Tax rate' }],
+    },
+  },
+  init(injector: Injector) {
+    pluginOptions = injector.get<ShippingExtensionsOptions>(PLUGIN_OPTIONS);
+  },
+  calculate: async (ctx, order, args) => {
```
martijnvdbrug, line 45: I think we need to handle the scenario where an `order.shippingAddress` isn't set yet. You could do `eligibleShippingMethods` query even for orders without shippingaddress right? I think it's fine to return $0.

Do you think we need this?

dalyathan: I think the other option we have is let the configured `OrderAddressToGeolocationConversionStrategy` handle this scenario, but I think returning 0 makes more sense than that @martijnvdbrug

packages/vendure-plugin-shipping-extensions/test/shipping-extensions.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+import {
+  POSTCODES_URL,
+  UKPostalCodeToGelocationConversionStrategy,
+} from '../src/strategies/uk-postalcode-to-geolocation-strategy';
+import { GeoLocation } from '../src/strategies/order-address-to-geolocation-strategy';
+import nock from 'nock';
+import { getDistanceBetweenPointsInKMs } from '../src/get-distance-between-points';
+import { CreateAddressInput } from '../../test/src/generated/shop-graphql';
+
+describe('Order export plugin', function () {
```
martijnvdbrug, line 36: You can remove the outter `describe` (this one even has a typo in it, probably my typo :laughing: )

martijnvdbrug (changes requested):
Impressive, nice work! Some minor change requested, but the setup and architecture is perfect :+1:

packages/vendure-plugin-shipping-extensions/README.md
dalyathan: Screenshot 2024-01-30 at 2 48 20 PM
 
I believe this will do it

martijnvdbrug (approved):
I approve!

PR #350, opened 2024-01-29
feat: support pushing and pulling of Picqer product data

nicoes:
# Description

I'd like to push the field fixedstockprice to Picqer but the Picqer Plugin doesn't allow this from a type perspective. Therefore adding support for that.


martijnvdbrug (approved):
Thanks!

PR #356, opened 2024-02-08
Feat: Accept Blue plugin progress

### LINKED ISSUE #346: Accept blue plugin
⚠ The WIP is on branch `feat/accept-blue-plugin`

We would like to duplicate the functionality of the Stripe Subscription plugin to also work with the [Accept Blue platform](https://docs.accept.blue/api/v2). A good starting point is to read [the README of the Stripe Subscription plugin](https://pinelab-plugins.com/plugin/vendure-plugin-stripe-subscription/).

In short, it allows you to define subscriptions (multiple) and an `amountDueNow` per order line. The `amountDueNow` is the amount that is actually paid during Vendure's checkout.

The Accept Blue flow is currently as follows:
1. Storefront calls `previewAcceptBlueSubscriptions()` query to preview what a subscription for a given variant will look like
2. Customer adds the variant to order
3. Customer calls `addPaymentToOrder` with its creditcard credentials
4. The plugin does: 1. getOrCreateCustomer, 2. getOrCreatePaymentMethod, 3. Creates a recurring schedule for the customer with the payment method id. This is a Accept Blue payment method. (This is already working)
5. What should still be implemented: A one time charge needs to be created, this is the actual payment of the order. The charge should be created with the saved payment method, not with the raw credit card credentials. It is not clear from the API yet how we can do that.

# What should be done
There is quite a lot of work still, so I propose to just pickup individual tasks and create PR's for those.

- [x] The one time charge needs to be implemented. This order should be settled. + e2e tests
- [x] Payments should also be possible with a tokenized payment method. Tokenization happens on the frontend: A customer enters its credit card credentials and Accept Blue gives us a token instead of card details, for security reasons. This way card details are never sent to the Vendure server. The token should be passed into the addPaymentToOrder mutation, similar to how the current credit card flow works.
- [x] `savedAcceptBluePaymentMethods` should be implemented. This returns a customers saved payment methods in Accept Blue. (We dont want to store the methods in Vendure). This call should both available on the admin-api and the shop-api
- [x] Payment should also be possible with [Check](https://docs.accept.blue/api/v2#tag/customers/paths/~1customers~1%7Bid%7D~1payment-methods/get). The plugin should then create a payment method with Check details instead of card details.

# What you need
(Credentials will be shared privately)
* The bulk of the setup is already available on `feat/accept-blue-plugin` branch
* The test environment dashboard can be found [here](https://sandbox.emeraldworldpayments.com/login). This is where you can view created customers, payments and schedules.
* You need a .env file locally, will share those details via Slack.
* `yarn start` runs the dev-server. That immediately creates a test order and a customer + payment method + recurring schedule. See dev-server.ts for details
* `http://localhost:3050/checkout` already has the hosted tokenization form implemented. It's just not sending the token to the backend yet.

dalyathan:
# Description

The changes in this implement and close #346
# Breaking changes

The price for the test `ProductVariant` with id `3` (or `T_3`) was changed from `2190.00` to `9.00` to make `Check/ ACH Charge` payment work with testing Check information in sandbox enviroment.(It requires that 'it(payments)..be under $15 for approval.')

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


packages/vendure-plugin-accept-blue/package.json
```diff
@@ -22,7 +22,7 @@
     "build": "rimraf dist && yarn generate && tsc",
     "start": "yarn ts-node test/dev-server.ts",
     "generate": "graphql-codegen",
-    "test": "vitest run --bail 1"
+    "test": "vitest run"
```
martijnvdbrug, line 25: Don't like the `--bail 1` option? (If it's not working, you might have to `yarn` in the root of the project).

Why I added it: Sometimes a test in the beginning fails, causing follow up tests to also fail. This made it hard for me to see what was the root cause.
The bail option stops any subsequent tests after one fails, so you know which one is the culprit

packages/vendure-plugin-accept-blue/src/api/accept-blue-common-resolvers.ts
```diff
@@ -65,19 +74,26 @@ export class AcceptBlueCommonResolver {
...
-    // return subscriptionsForOrderLine.map((s) => ({
-    //   ...s,
-    //   variantId: orderLine.productVariant.id,
-    // }));
-    throw new Error('Not implemented');
+@Resolver('AcceptBluePaymentMethod')
+export class SavedPaymentMethodsResolver {
+  @ResolveField()
+  __resolveType(value: any): string {
+    return value.payment_method_type === 'card'
```
martijnvdbrug, line 95: You should be able to put this in the same `AcceptBlueCommonResolver` above, so that we don't need a separate `SavedPaymentMethodsResolver`

Keep in mind that you need a specific order for this to work:
```ts
// This works
@ResolveField('__resolveType')
@Resolver('AcceptBluePaymentMethod')

// While this doesn't
@Resolver('AcceptBluePaymentMethod')
@ResolveField('__resolveType')
```

packages/vendure-plugin-accept-blue/src/api/accept-blue-common-resolvers.ts
```diff
@@ -65,19 +74,26 @@ export class AcceptBlueCommonResolver {
     @Ctx() ctx: RequestContext,
```
martijnvdbrug, line 74: Not part of this PR, but it says `stripeSubscriptions` instead of `acceptBlueSubscriptions`

packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
```diff
@@ -68,31 +79,186 @@ export class AcceptBlueService {
...
+      order.customer.emailAddress
+    );
+    if (!order.customer?.customFields?.activeBlueCustomerId) {
+      await this.customerService.update(ctx, {
+        id: order.customer?.id,
+        customFields: { activeBlueCustomerId: customer.id },
+      });
+    }
+    // Save payment method
+    const methods = await client.getPaymentMethods(
```
martijnvdbrug, line 123: Comment doesn't seem correct, this only checks if a payment method with the token already exists right?

Also: Is a payment method already created when we have a tokenized card? I don't fully understand how it works yet.

packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
```diff
@@ -68,31 +79,186 @@ export class AcceptBlueService {
...
+      order.customer.customFields.activeBlueCustomerId
+    );
+    const paymentMethod = methods.find((method) =>
+      isSameCard(ccDetails, method as AcceptBlueCardPaymentMethod)
+    );
+    if (!paymentMethod) {
+      throw new UserInputError(
+        'No Saved Credit Card with this toknized nonce code'
```
martijnvdbrug, line 131: Typo `toknized`

packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
```diff
@@ -68,31 +79,186 @@ export class AcceptBlueService {
...
+      ccDetails
     );
   }
 
+  async payWithCheck(
+    ctx: RequestContext,
+    order: Order,
+    amount: number,
+    client: AcceptBlueClient,
+    ccDetails: CheckPaymentInput
```
martijnvdbrug, line 149: `cc` stands for creditcard, so in this case we can rename it to `checkDetails` or something?

Same goes for `payWithToken` input above

packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
```diff
@@ -68,31 +79,186 @@ export class AcceptBlueService {
...
+  ): Promise<HandlePaymentResult> {
+    if (!order.customer) {
+      throw new UserInputError(`Order must have a customer`);
+    }
+    const customer = await client.getOrCreateCustomer(
+      order.customer.emailAddress
+    );
+    if (!order.customer?.customFields?.activeBlueCustomerId) {
```
martijnvdbrug, line 157: Typo I guess? Shouldn't this be `acceptBlueCustomerId` instead of activeBlue...?

packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
```diff
@@ -68,31 +79,186 @@ export class AcceptBlueService {
...
+  }
+
+  async payWithCheckPaymentMethodDetails(
+    ctx: RequestContext,
+    order: Order,
+    amountDueNow: number,
+    client: AcceptBlueClient,
+    paymentMethodId: number,
+    ccDetails: CheckPaymentInput
+  ): Promise<HandlePaymentResult> {
```
martijnvdbrug, line 229: How come we accept the `ccDetails` AND `paymentMethodId`? Can't we follow this flow for all payment types (check, card and tokenized)?

Get or Create customer > get Or Create Payment method for customer > **Create payment with saved payment method** (not passing around the details)

packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
```diff
@@ -68,31 +79,186 @@ export class AcceptBlueService {
...
+
+  async payWithCardPaymentMethodDetails(
+    ctx: RequestContext,
+    order: Order,
+    amountDueNow: number,
+    client: AcceptBlueClient,
+    paymentMethodId: number,
+    ccDetails: CreditCardPaymentInput | TokenPaymentMethodInput
+  ): Promise<HandlePaymentResult> {
+    const recurringSchedules =
```
martijnvdbrug, line 186: I see that you are only passing the `paymentMethodId` in here to create the recurring schedules here. This function now does 2 things: create recurring schedules and create a one time payment.

To avoid too many function layers (too many function to function to function calls), it's better to use this flow in the upper `payWithCreditCard()` function:

```ts
await createRecurringSchedule(ctx, order, paymentMethod);
// Preferably use paymentMethodId, if not possible pass ccDetails
await createOneTimePayment(ctx, order, amountDueNow, paymentMethodId)
```
This way you have 2 clear separate functions `createRecurringSchedule` and `createOneTimePayment`, instead of `payWithCardPaymentMethodDetails` and `payWithCreditCard` (they sound pretty similar right?)

It's usually a sign things can be structured better if you are passing the same arguments (paymentMethodId, client) down multiple layers of functions.

martijnvdbrug (changes requested):
Changes and explanations required :-) Nice work so far though

packages/vendure-plugin-accept-blue/src/api/accept-blue-handler.ts
```diff
@@ -59,19 +61,19 @@ export const acceptBluePaymentHandler = new PaymentMethodHandler({
     args,
     metadata
   ): Promise<CreatePaymentResult> {
-    const ccMetadata = metadata as CreditCardPaymentInput;
+    const ccMetadata = metadata as PaymentInput;
     const client = new AcceptBlueClient(args.apiKey, args.pin);
-    if ((metadata as SavedPaymentMethodInput).paymentMethodId) {
-      const { paymentMethodId } = metadata as SavedPaymentMethodInput;
-      const result = await service.payWithSavedPaymentMethod(
+    const deNormalizedAmountValue = amount / 100;
```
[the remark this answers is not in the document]
dalyathan: Makes sense

packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
dalyathan: We use the `paymentMethodId` to create recurring schedules and we use `ccDetails` to create a one time charge. What do you think about this approach @martijnvdbrug ?

martijnvdbrug: @dalyathan Thats fine for now! I was just hoping we could do a one time charge with saved payment methods, so that customers could in the future check out with just a saved payment method.

martijnvdbrug (approved):
Nice work on the small refactorings after PR comments :+1:

PR #359, opened 2024-02-11
feat(vendure-order-client): modified the active customer query to get…

ZackNew:
- updated the activeCustomer query to get default billing and shipping addresses


martijnvdbrug:
Closing this, as vendure-order-client is no more. We use Codegen and Vue Composables in our client projects now

PR #364, opened 2024-02-14
Invoice UI Improvements

### LINKED ISSUE #352: Invoices plugin improvements
After using the invoices plugin for one of our clients, I think we can improve the UI a bit:

* Can we remove the outer according that says `Settings`? This was used because it was shown together with a list of invoices. Now, in V2 it's a dedicated settings page, so it doesn't make sense to have it collapsable
* Can we put the form fields and buttons in a container, to make it more consistent with Vendure's styling. See the attached screenshot  of Global Settings. All detail screens have the same basic layout
   * Move `save` button to the top right and rename to `update`
* Can we make the order-code optional? If no order code is given, we should get the latest placed order.
* Remove the Preview tooltip. It's description isn't accurate anymore. We can rename the button to 'Preview template'
* We should keep the preview button close to the template and close to the orderCode field, so that it's clear those are all related to eachother.

Image

Image

dalyathan:
# Description

The changes in this PR implement and close #352 
# Breaking changes

# Screenshots

Screenshot 2024-02-14 at 3 25 12 PM

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:

### DIFF
--- a/packages/vendure-plugin-invoices/src/services/invoice.service.ts
+++ b/packages/vendure-plugin-invoices/src/services/invoice.service.ts
@@ -40,6 +40,7 @@ import { createTempFile } from '../util/file.util';
 import { reverseOrderTotals } from '../util/order-calculations';
 import { InvoiceCreatedEvent } from './invoice-created-event';
 import { QueryDeepPartialEntity } from 'typeorm/query-builder/QueryPartialEntity';
+import { SortOrder } from '@vendure/common/lib/generated-shop-types';
 
 interface DownloadInput {
   customerEmail: string;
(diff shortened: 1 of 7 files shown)


packages/vendure-plugin-invoices/src/services/invoice.service.ts
```diff
@@ -264,9 +265,24 @@ export class InvoiceService implements OnModuleInit, OnApplicationBootstrap {
...
+        await this.orderService.findAll(
+          ctx,
+          {
+            take: 1,
+            sort: { createdAt: SortOrder.DESC },
+          },
+          []
+        )
+      )?.items[0].id;
+      order = await this.orderService.findOne(ctx, orderId);
```
martijnvdbrug, line 284: 1. Can we just get the latest placed order?
2. Why not use `items[0]` instead of using it's ID to refetch an order?

martijnvdbrug (changes requested):
Looks good, minor changes requested :-)

packages/vendure-plugin-invoices/src/services/invoice.service.ts
dalyathan: `OrderService.findOne` includes more relations, translates entities and also applies prices to the `ProductVariant`'s in the Order. You don't think this is a good approach @martijnvdbrug ?

martijnvdbrug (approved):
Nice work, thanks!

PR #373, opened 2024-02-26
RequestContext should also be an argument to ShippingExtensionsOptions.weightCalculationFunction

dalyathan:
# Description

The changes in this PR adds `RequestContext` as an additional argument to `ShippingExtensionsOptions.weightCalculationFunction`
# Breaking changes

We will need to pass `RequestContext` when calling `ShippingExtensionsOptions.weightCalculationFunction`
# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:

## Summary by CodeRabbit

- **New Features**
	- Enhanced shipping weight calculation by including request context for more accurate and context-aware calculations.

### DIFF
--- a/packages/vendure-plugin-shipping-extensions/CHANGELOG.md
+++ b/packages/vendure-plugin-shipping-extensions/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.5.1 (2024-02-26)
+
+- Include `RequestContext` as an additional argument to `ShippingExtensionsOptions.weightCalculationFunction`
+
 # 2.5.0 (2024-02-19)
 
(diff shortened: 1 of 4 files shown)


martijnvdbrug (approved):
Yep, perfect, ctx as first argument, nice. Next time this could have been a minor update, but it's fine since we are the only ones using it right now.

* patch = non-breaking bugfix. Consumers should always update this. No api and interface changes allowed, e.g. no new fields in `Plugin.init()` method.
* minor = non-breaking new feature. Can be small like this, but it's not necessary for consumers to update. E.g. they don't have to use `ctx` if they are not using it now
* major: all other breaking stuff

PR #397, opened 2024-04-08
fix: remove fallback shipping address for billing address

nicoes:
We've changed the front-end check-out to make filling out the shipping address the default instead of making the billing address the default. It occurred to us that the shipping address is always filled in the billing address. Which doesn't make sense for us (anymore). From a Picqer API perspective the invoiceaddress is not required. So I think it's better to remove the fallback.

https://picqer.com/en/api/orders

### DIFF
--- a/packages/vendure-plugin-picqer/CHANGELOG.md
+++ b/packages/vendure-plugin-picqer/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.4.4 (2024-04-09)
+
+- Don't fallback to shipping address when passing billing address to Picqer
+
 # 2.4.3 (2024-04-03)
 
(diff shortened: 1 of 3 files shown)


packages/vendure-plugin-picqer/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.4.4 (2024-04-09)
+
+Don't fall back to shipping address when passing billing address to Picqer
```
martijnvdbrug, line 3:
```suggestion
- Don't fall back to shipping address when passing billing address to Picqer
```

martijnvdbrug (approved):
Thanks!

PR #402, opened 2024-04-12
Cancelled Order Invoice Updates

### LINKED ISSUE #401: When an order is cancelled, the invoice plugin should create a credit invoice!
Currently credit invoices are only generated when someone clicks 'regenerate invoice' in the admin. 2 things need to happen:

1. Whenever an order is cancelled, a credit invoice should be generated
2. If an order is cancelled, it should not be possible to regenerate invoices for that order!

dalyathan:
# Description

The changes in this PR implement and close #401 
# Breaking changes

# Screenshots

Screenshot 2024-04-12 at 12 02 19 PM

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


packages/vendure-plugin-invoices/src/credit-invoice-generation.process.ts
```diff
@@ -0,0 +1,29 @@
...
+
+let invoiceService: InvoiceService;
+
+export const creditInvoiceGenerationProcess: OrderProcess<'Cancelled'> = {
+  init(injector) {
+    invoiceService = injector.get(InvoiceService);
+  },
+
+  async onTransitionStart(_, toState, data) {
+    const previousInvoiceForOrder =
```
martijnvdbrug, line 12: We might need to account for the situation where an order doesn't have a previous invoice. I imagine this:

1. When a previous invoice isnt there, we log an error but dont throw
2. When invoice generation fails its good to throw an error, because this should block cancellation

packages/vendure-plugin-invoices/src/ui/invoices-nav.module.ts
```diff
@@ -61,6 +65,16 @@ import { ApolloCache } from '@apollo/client/cache';
             (event.target as HTMLButtonElement).disabled = false;
           });
       },
+      buttonState: (context) => {
+        const orderId = context.route.snapshot.params['id'];
```
martijnvdbrug, line 69: Minor possible improvement: i remember something about the entity (the order in this case) is available on the context object. If thats the case we dont need to do an extra query

martijnvdbrug (changes requested):
Super nice cleanup and new feature! Ive spotted 2 minor things, but other than that its a very nice and clean PR.

Can you create a GH task on the pinelab board to update the invoice plugin for CG ?

packages/vendure-plugin-invoices/src/ui/invoices-nav.module.ts
dalyathan: I think `ActionBarContext` only has the following fields: 
```ts
{
    route: ActivatedRoute;
    injector: Injector;
    dataService: DataService;
    notificationService: NotificationService;
}
```

martijnvdbrug (changes requested):
So, I thought of something that should actually still be in the PR:

The invoice generation should be done in the worker instance, so the cancellation should send a job to the job queue. The job handler now assumes that every job means that we create-and-save a credit invoice + a new invoice though, so we need to change some things:

1. An argument should be added to the job, indicating that only a creditInvoice should be created:
```diff
// line 54
- jobQueue: JobQueue<{ channelToken: string; orderCode: string }> | undefined;
+ jobQueue: JobQueue<{ channelToken: string; orderCode: string, creditInvoiceOnly: boolean }> | undefined;
```
2. The process function should differentiate between `createAndSaveInvoice` and 
```ts
// line 74
  async onModuleInit(): Promise<void> {
    this.jobQueue = await this.jobService.createQueue({
      name: 'generate-invoice',
      process: async (job) =>
        if (job.data.creditInvoiceOnly) {
          this.createAndSaveCreditInvoice()
        } else {
          // create and save invoice
        }
    });
  }
```

martijnvdbrug (changes requested):
Nice work already. I'm making some changes locally, so **no action needed**, but here is the feedback so far:

1. Make sure eventbus code is always wrapped in a try/catch. Throwing an error in a `.subscribe` wil kill the entire node process!
2. Whenever you have functions with similar names, it's usually an indication code is not properly grouped. I.e. `createCreditInvoiceForCancelledOrder` calls `createAndSaveCreditInvoice`. The functions look very similar, so it might be confusing for future devs to know why these 2 functions are not 1 single function. I solved it in this specific by handling cancelled orders also in the `createAndSaveInvoice` function, but allow specifying a creditOnly argument.

martijnvdbrug:
Also added e2e tests for cancelling orders

martijnvdbrug (approved):

PR #416, opened 2024-05-07
Vendure scripts

### LINKED ISSUE #415: Assign Products to Channel Script
For both Wholesale and the new Fabrix channel for CG, we'd need to assign products to other channels. To be able to reuse things like this, we should create a publishable package in the plugins repository: "@pinelab/vendure-scripts"

It should expose functions that can be used on Vendure projects. To start, we'd need 2 functions:

1. `assignAllProductsToChannel`. A helper function that assigns all products of channel X to channel Y, including:
   * Facets and facetvalues connected to the products
   * Assets connected to the products
   * Anything else we might be missing?
2. `assignProductsToChannel([1,3,4,5,6])` (list of product ids to assing to channel).

Some acceptance requirements:
* These scripts will be run locally mostly, so no need to use jobqueue
* It should be scalable for large projects: We should not load ALL products then assign all of them, but rather in batches.
* Script should be idempotent: running the script multiple times should not break anything. It should just assign 1 product to the new channel once.

## Example code

I imagine we have something like this. 

```ts
import { assignAllProductsToChannel } from '@pinelab/vendure-scripts';

// Consumers load their Vendure app here first, before using our script, because we need them to load their .env and connect to their db etc.
await bootstrap();

await assignAllProductsToChannel(ctx, {
   sourceChannelId: 1,
   targetChannelId: 2,
});
```

dalyathan:
# Description

The changes in this PR implement and close #415 
# Breaking changes

The product `Laptop` will have facets in the test environment, not sure it will be breaking tho...
# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


packages/test/src/admin-utils.ts
```diff
@@ -189,3 +196,33 @@ export async function createPromotion(
   });
   return createPromotion as Promotion;
 }
+
+export async function getSuperadminContextInChannel(
```
martijnvdbrug, line 200: I see you are using this helper in the source code of the assign products script. In that case, can you move it to `packages/util` instead of `packages/test`? `Packages/test` should only contain helpers that are used in tests

packages/vendure-scripts/src/assign-these-prodcuts-to-channel.ts
```diff
@@ -0,0 +1,62 @@
...
+  Injector,
+  Product,
+  ProductVariantService,
+  RequestContext,
+  SearchService,
+  TransactionalConnection,
+} from '@vendure/core';
+import { getSuperadminContextInChannel } from '../../test/src/admin-utils';
+
+export async function assignTheseProductsToChannel(
```
martijnvdbrug, line 15: Can you add some JS doc here, stating that it commits all products in a single transaction, thus that it should be used with batches of products?

packages/vendure-scripts/README.md
```diff
@@ -0,0 +1,17 @@
+# Vendure Script
+
+## [Official documentation here](https://pinelab-plugins.com/plugin/vendure-scripts)
+
+This 'plugin' contains generic scripts that can be used with `Vendure` projects.
+
+## Getting started
```
martijnvdbrug, line 7: Can you add a separate section for `Assign all products to channel`, with a short description about what this script does and what its for?

packages/vendure-scripts/src/assign-these-prodcuts-to-channel.ts
```diff
@@ -0,0 +1,62 @@
...
+      productVariantIds: variantIds,
+    }),
+    //we will then assign the product assets to the target channels
+    assetService.assignToChannel(ctx, { assetIds, channelId: targetChannelId }),
+    //we will then assign the facets with the associated values
+    facetService.assignFacetsToChannel(ctx, {
+      channelId: targetChannelId,
+      facetIds,
+    }),
+    searchService.reindex(superadminContextInTargetChannel),
```
martijnvdbrug, line 59: All these calls need to be awaited, even though they are inside a Promise.all. When you don't await them, they sort of dissapear in the background, and thus will not be awaited by the promise.all

martijnvdbrug (changes requested):
Nice work! Some changes requested. Haven't reviewed it all yet though..

packages/vendure-scripts/src/assign-these-prodcuts-to-channel.ts
```diff
@@ -0,0 +1,62 @@
...
+    )
+    .flat();
+  const assetIds = products
+    .map((product) => (product.assets ?? []).map((asset) => asset.id))
+    .flat();
+  const variantIds = products
+    .map((product) => (product.variants ?? []).map((variant) => variant.id))
+    .flat();
+  const targetChannel = await channelService.findOne(ctx, targetChannelId);
+  const superadminContextInTargetChannel = await getSuperadminContextInChannel(
```
martijnvdbrug, line 42: Can't we use the request context that the consumer gives us, instead of creating our own super admin credentials?

packages/vendure-scripts/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "author": "Surafel Tariku <surafelmelese09@gmail.com>",
+  "homepage": "https://pinelab-plugins.com/",
+  "repository": "https://github.com/Pinelab-studio/pinelab-vendure-plugins",
+  "license": "MIT",
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
```
martijnvdbrug, line 15: The paths are actually `dist/vendure-scripts/src/index.js` (and .d.ts) because we import utils from /test and /utils

martijnvdbrug (changes requested):
2 more minor changes I missed in the previous review

packages/test/src/admin-utils.ts
dalyathan: It is actually used by both tests and src code. Where should it belong in that case? @martijnvdbrug

packages/vendure-scripts/src/assign-these-prodcuts-to-channel.ts
dalyathan: I think is it best if we don't have batch level transactions. I will update the code to have the transaction started before all Products are processed and committed when we are done with importing all batches

packages/vendure-scripts/src/assign-these-prodcuts-to-channel.ts
dalyathan: I think the configured SearchPlugin uses the `ctx.channelId` when reindexing. See [this](https://github.com/vendure-ecommerce/vendure/blob/0b1dcb7b03ca127ac8e63540d61d13fbcc02ff9f/packages/core/src/plugin/default-search-plugin/indexer/indexer.controller.ts#L332) for the DefaultSearchPlugin.

packages/test/src/admin-utils.ts
martijnvdbrug: In that case I would say `packages/util`. 

The main reason that the folder ends up in the compiled `dist`:
* `dist/util/src/...`
* `dist/vendure-scripts/src/..`

While its already not too great that we have `src` in dist, it will confuse other devs even more if we have a folder named `test` in the dist :-)

packages/vendure-scripts/src/assign-these-prodcuts-to-channel.ts
martijnvdbrug: That makes sense :+1

packages/vendure-scripts/src/assign-these-prodcuts-to-channel.ts
martijnvdbrug: Goood find :ok_hand: . Can we make it so that we only pass a custom Ctx to the `reindex()` call, and use the given ctx for the other calls? The reason for this is that this script **might** trigger customer project in the consumer code like eventlisteners. That custom code might give unexpected results with a Ctx constructed by us.

packages/vendure-scripts/src/assign-products-to-channel.ts
```diff
@@ -0,0 +1,41 @@
+import {
+  ID,
+  Injector,
+  Product,
+  RequestContext,
+  TransactionalConnection,
+} from '@vendure/core';
+import { In } from 'typeorm';
+import { assignTheseProductsToChannel } from './assign-these-prodcuts-to-channel';
```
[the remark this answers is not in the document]
martijnvdbrug: @dalyathan there seems to be a typo in the filename

packages/vendure-scripts/src/assign-products-to-channel.ts
```diff
@@ -0,0 +1,41 @@
...
+  const conn = injector.get(TransactionalConnection);
+  await conn.startTransaction(ctx);
+  do {
+    // get products of the source channel with id productIds
+    products = await getProductsDeep(
+      ctx,
+      injector,
+      { id: In(productIds) },
+      batch,
+      totalCount
```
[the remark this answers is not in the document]
martijnvdbrug: This function is already inside a do/while, where products are fetched in batches

packages/vendure-scripts/src/assign-these-prodcuts-to-channel.ts
dalyathan: I think that is the case in the current implementation @martijnvdbrug

martijnvdbrug (approved):
Nice work!

PR #429, opened 2024-05-27
Shipmate Plugin

### LINKED ISSUE #400: ShipMate plugin
Quite a big and complex task. Don't hestitate to ask. Maybe we can do some iterations where you set up the desired structure and interfaces first?

For the CG project we would like to integrate with [ShipMate](https://www.shipmate.co.uk/). ShipMate is a platform that helps you ship orders. It works like this:
1. Vendure sends an order to ShipMate
2. CG employee print labels from ShipMate to add to the physical order
3. CG Employees add the labels to the physical order and bringe the to a shipping point or dropoff point or w/e.

## Requirements
You probably need to create an account to be able to test the intergration. If you need to supply billing information you can ask James for a developer account
* When an order is placed, the order should be sent to ShipMate.
* When an order is updated, we should update the same order in Shipmate
* The plugin should be channel aware, so we should be able to configure API keys in the Admin. See MyParcel plugin for reference
* On Vendure startup, Vendure should automatically configure webhooks in ShipMate, so that Vendure will receive updates from ShipMate
* When an order in ShipMate sends TRACKING_COLLECTED via webhook, we should update the Vendure order to Shipped
* When an order in ShipMate sends TRACKING_DELIVERED, we should update the Vendure Order to Delivered
* We dont support partial shipping for now

We should send the following fields to Shipmate. This is a plugin, so we cannot directly depend on custom fields like addressLine3:
* All default Vendure address fields like Street City etc
* For the CG project we also need to send `weight` to ShipMate
* The CG project also requires customFields.addressLine3 to be send to ShipMate.
To solve weight and address customFields, we can maybe allow some strategies in the plugin, that allow the consumer to send extra fields to ShipMate:
```ts
// example
Plugin.init({
   additionalParcelFieldsFn: async (injector, orderLine) => {
        // do your weight magic in the CG project, not in the plugin
        return {
           weight: 300
        }
    },
   additionalAddressFields: async (injector, orderAddress) => {
        // Return fields should be typed based on the ShipMate allowed address fields. 
        return {
           line_1: orderAddress.customFields.addressLine3
        }
    },
})
```

## Additional context
API Docs: https://www.shipmate.co.uk/guides/api

dalyathan:
# Description
The changes in this PR implement and close #400 

# Breaking changes

# Screenshots

Screenshot 2024-05-27 at 7 17 05 PM

Screenshot 2024-05-27 at 7 11 43 PM

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


packages/vendure-plugin-shipmate/test/shipmate.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+import nock from 'nock';
+
+import { VendureShipmatePlugin } from '../src/shipmate.plugin';
+import { newShipment } from './test-helpers';
+import { createSettledOrder } from '../../test/src/shop-utils';
+import { testPaymentMethod } from '../../test/src/test-payment-method';
+import { getSuperadminContext } from '@vendure/testing/lib/utils/get-superadmin-context';
+import { ShipmateConfigService } from '../src/api/shipmate-config.service';
+import axios from 'axios';
+import { TrackingEventPayload } from '../src/types';
```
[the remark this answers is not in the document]
dalyathan: we need axios not as a type but as a value

packages/vendure-plugin-shipmate/test/shipmate.spec.ts
```diff
@@ -0,0 +1,150 @@
+import {
+  AutoIncrementIdStrategy,
+  DefaultLogger,
+  LogLevel,
+  OrderService,
+  RequestContext,
+  mergeConfig,
+} from '@vendure/core';
```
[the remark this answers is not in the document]
dalyathan: we need `DefaultLogger`, `LogLevel` and `OrderService` as a value

packages/vendure-plugin-shipmate/src/api/shipmate-config.entity.ts
```diff
@@ -0,0 +1,24 @@
+import { Column, Entity } from 'typeorm';
+import { DeepPartial, VendureEntity } from '@vendure/core';
+
+@Entity()
+export class ShipmateConfigEntity extends VendureEntity {
+  constructor(input?: DeepPartial<ShipmateConfigEntity>) {
+    super(input);
+  }
```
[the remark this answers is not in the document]
dalyathan: we need this constructor. otherwise `plugin.ts` file won't compile

packages/vendure-plugin-shipmate/src/api/shipmate-config.service.ts
```diff
@@ -0,0 +1,63 @@
+import { Injectable } from '@nestjs/common';
+import { RequestContext, TransactionalConnection } from '@vendure/core';
```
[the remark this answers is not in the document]
dalyathan: Nestjs Providers to be injected in a constructor of another Provider need to be imported as a value

dalyathan:
It is better to store the `webhookAuthTokens` in a separate table rather than have them in a commas separated `varchar` column. I will work on this one

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
```diff
@@ -0,0 +1,271 @@
...
+        shipmateConfig.password
+      );
+      this.tokens.set(shipmateConfig.apiKey, shipmateTokenForTheCurrentChannel);
+    }
+    this.httpService.axiosRef.defaults.headers.common[
+      SHIPMATE_TOKEN_HEADER_KEY
+    ] = shipmateTokenForTheCurrentChannel;
+    this.httpService.axiosRef.defaults.headers.common[
+      SHIPMATE_API_KEY_HEADER_KEY
+    ] = shipmateConfig.apiKey;
```
martijnvdbrug, line 85: This is not concurrency-safe: When 2 channels simultatiously do a shipmate action, both will set the headers, but they set headers on the same `this.httpService.axiosRef` instance, so only the last one will be persisted. This means that the first channel will be making calls with the headers of the second channel. This would create orders of channel 1 in the shipmate account of channel 2. 

A good practice is to get a client ( or axios instance) on every action, based on the CTX. Similar to how we do that in the accept-blue pllugin.

1. Axios instance example (pseudo code)
```ts
getClient(ctx: RequestContext): Promise<AxiosInstance> {
   const client = axios.create();
   client.setHeaders.... etc
}
```
2. Even better: Create a separate client class that abstracts the API calls, and does HTTP validations etc.
```ts
// separate client class
export class ShipmateClient {
    constructor(username, password) {
        // create client with username password
        }

        // pass in the already transformed shipment object ready for Shipmate API
        createShipment(shipment) {
            // Get API token based on username password
            // Do the create call
            // Validate response and throw error if it didnt work
        }
}

// In this service we can then do:
getClient(ctx: RequestContext): Promise<ShipmateClient> {
    const config = getConfig(ctx);
    return new ShipmateClient(config.username, config.password);
 }
```

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
```diff
@@ -0,0 +1,271 @@
...
+          password: shipmatePassword,
+        }
+      )
+    );
+    if (response.data.data?.token) {
+      Logger.info('Successfully authenticated with Shipmate API', loggerCtx);
+      return response.data.data.token;
+    } else {
+      Logger.error(response.data.message, loggerCtx);
+      throw new UserInputError(response.data.message);
```
martijnvdbrug, line 106: This one would go in the ShipMate client class

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
```diff
@@ -0,0 +1,271 @@
...
+        .status(HttpStatus.CREATED)
+        .json({ message: `Order marked as Delivered successfully` });
+    } else {
+      Logger.info(
+        `No configured handler for event "${payload.event}"`,
+        loggerCtx
+      );
+    }
+    await this.connection.commitOpenTransaction(ctx);
+    return;
```
martijnvdbrug, line 171: `res` is pretty `controller` specific, but a service should only contain business logic, and not really be aware of HTTP or Graphql. A better approach is to throw an error here (UserInputError, ForbiddenError etc), and let the controller decide how this should be returned as HTTP error.

Considering this is a webhook, it's also fine to just differentiate between Error or returning. The different HTTP return statusses aren't used by the shipmate webhook i think? They only care about OK or not, and retry if not OK.

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
```diff
@@ -0,0 +1,271 @@
...
+    private eventBus: EventBus,
+    private connection: TransactionalConnection,
+    private entityHydrator: EntityHydrator
+  ) {}
+
+  async onModuleInit(): Promise<void> {
+    this.httpService.axiosRef.defaults.headers.common['Accept'] =
+      'application/json';
+    this.httpService.axiosRef.defaults.headers.common['Content-Type'] =
+      'application/json';
```
martijnvdbrug, line 54: These header settings should go in the ShipmateClient

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
```diff
@@ -0,0 +1,271 @@
...
+      Logger.info('Successfully authenticated with Shipmate API', loggerCtx);
+      return response.data.data.token;
+    } else {
+      Logger.error(response.data.message, loggerCtx);
+      throw new UserInputError(response.data.message);
+    }
+  }
+
+  async createShipment(ctx: RequestContext, order: Order) {
+    const shipmateReference = generatePublicId();
```
martijnvdbrug, line 111: I propose we also use the order code as shipmate reference. The unique ID is only for our internal reference, and we dont save it anywhere else, so it isn't really a reference to anything. I guess this is used when you would create multiple shipments per order

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
```diff
@@ -0,0 +1,271 @@
...
+    const payload = parseOrder(order, shipmateReference);
+    try {
+      const result = await firstValueFrom(
```
martijnvdbrug, line 114: Do we really need `firstValueFrom`? Can we use promises? API calls are usually just one-off calls, not data streams like you would have in the frontend

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
```diff
@@ -0,0 +1,271 @@
...
+    }
+  }
+
+  async updateOrderState(payload: EventPayload, res: Response): Promise<void> {
+    const ctx = await this.createCtx(payload.auth_token);
+    if (!ctx) {
+      const message = `No registered ShipmentConfigEntity with this auth_token`;
+      Logger.error(message, loggerCtx);
+      res.status(HttpStatus.BAD_REQUEST).json({ message });
+      return;
```
martijnvdbrug, line 137: This should throw an Error, and let the controller handle HTTP response statusses. Logging is indeed a good practice though.

martijnvdbrug (changes requested):
Nice work already. Some changes requested. Don't worry, its very rare to get big things like these right the first time.

* The main issue is the Shipmate client, we need that one to be concurrency safe
* I do like the dedicated config service, definitely a good practice!
* Can we rename `shipmate.admin.graphql.ts` to `api-extensions.ts`? I do like your naming convention better, but it's worth more to stay in line with the Vendure (and other plugins) conventions, where the files are all named api-extensions.ts
* Can you add the README and CHANGELOG contents? 
* Feel free to add yourself as author of the plugin if you wish :) 
* Nice work on the mappers (parseorder etc), perfect!

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
dalyathan: Since creating the axios instance depends on the credentials provided by `ShipmateConfigService.getConfig`, it will only be encapsulating `post` calls. Do you think this is worth having its own class @martijnvdbrug ?

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
dalyathan: I used that because the `Observable.toPromise` call is deprecated. But `axios.post` returns a promise so this won't be used once changes are made

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
dalyathan: I guess we can make the `ShipmateClient` it self a service class

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
martijnvdbrug: I do see you are checking and throwing  response.data.message, this is Shipmate specific response data. This would be an argument to create a separate Client class though

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
martijnvdbrug: Why are we manually starting and commiting transactions?

packages/vendure-plugin-shipmate/src/api/shipmate.service.ts
dalyathan: I checked shipmate's api we don't really need to return 201. Any success code in the 200 series will do

martijnvdbrug (approved):
* Changed client to accept username/password and APIKey
* Headers shouldn't be passed around, they are Shipmate specific, so they dont really belong in the service, but rather in the client; It's the clients responsibility to use tokens and API keys as headers
* `getShipmentToken` and `getShipmateAuthHeaders` both belong inside the client. The fact that username/password need to be traded for an auth token is Shipmate specific, not really business logic.
* Downside is that we recreate shipmate tokens more often. But I think this is a small trade off compared to encapsulation. You could always create a static `HashMap<ctx, token>` inside the client if needed.
* It's better to send to Shipmate via the worker instance. It's vital that shipments end up in Shipmate, and we want to retry if Shipmate (or some network request) temporarily fails.

You can see the changes in here: https://github.com/Pinelab-studio/pinelab-vendure-plugins/pull/429/commits/6697ffadefb4bf2b36ccba5169364f5735accd0a

martijnvdbrug (changes requested):
Final changes requested. Will pick this up myself tomorrow morning:

* ~~Webhook tokens back to inline array of seperate entity? Awaiting Surafels answer on this~~
* Add to readme: installation, admin recompile, create config via Admin + create webhooks

martijnvdbrug (approved):
finalized!

PR #432, opened 2024-05-31
Assigning Customers and Orders

### LINKED ISSUE #430: Assign customers and orders to channel
Similar to the assign products script, we would also like 2 additional scripts:

* Assign all customers from channel X to channel Y
* Assign all orders from channel X to channel Y

I found this in Vendure source. It seems to be a scalable way to assign entities to other channels. What do you think? Or are we already using this:

```ts
      return app.get(DataSource)
      .getRepository(Customer)
      .createQueryBuilder()
      .relation('channels')
      .of(id)
      .add(targetChannel.id).catch((e) => {
        console.error(`Error assigning customer ${id} to channel ${targetChannel.token}`, e);
      });
```

dalyathan:
# Description

The changes in this PR implement and close #430 

# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


martijnvdbrug (approved):
Nice work and cleanup!

PR #463, opened 2024-07-08
feat(accept-blue): allow test mode to be set at payment handler level

mschipperheyn:
# Description

This allows you to set test mode for the accept blue plugin at the payment handler level. This is desirable in multi vendor scenario's where each vendor has an individual accept blue configuration

# Breaking changes

No, the .env variable for testMode is maintained and acts as a global override.

# Screenshots

# Checklist

📌 Always:
- [ x] I have set a clear title
- [ x] My PR is small and contains a single feature
- [x ] I have [checked my own PR](## "Fix typo's and remove unused or commented out code")

👍 Most of the time:
- [ x] I have added or updated test cases for important functionality
- [x ] I have updated the README if needed

📦 For publishable packages:
- [ x] I have [increased the version number](## "After increasing the version to the next patch/minor/major, the package will be published automatically after merge") in `package.json`
- [ x] I have added my changes to the `CHANGELOG.md` [See this example](https://github.com/Pinelab-studio/pinelab-vendure-plugins/blob/main/packages/vendure-plugin-invoices/CHANGELOG.md)

### DIFF
--- a/packages/vendure-plugin-accept-blue/CHANGELOG.md
+++ b/packages/vendure-plugin-accept-blue/CHANGELOG.md
@@ -1,6 +1,7 @@
-# 1.4.1 (2024-07-08)
+# 1.5.0 (2024-07-08)
 
-- Allow test mode to be set at payment method level
+- Allow test mode to be set at payment method level.
+- Removed the usage of env var ACCEPT_BLUE_TEST_MODE for test mode. Use the setting on payment method.
 
 # 1.4.0 (2024-06-21)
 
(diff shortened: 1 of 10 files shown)


martijnvdbrug (approved):
Thanks @mschipperheyn ! 

For the sake of speed I took the liberty to make some minor changes in [this commit](https://github.com/Pinelab-studio/pinelab-vendure-plugins/pull/463/commits/0b0ac56462b6b13fda4569302aa6ae147246098e) instead of leaving a review:
1. I bumped the version to a new minor instead of a patch, 1.5.0, because it introduces a new field in the payment method handler
2. I removed the usage of the test env var completely to avoid confusion, your approach is better. (Other devs will set the env to true, and admins will start asking questions why AC is used in test mode when it isn't checked in their env vars. And admins don't read readme's unfortunately :laughing: )

PR #465, opened 2024-07-09
zone aware shipping calculator

dalyathan:
# Description

The changes in this PR implement zone aware shipping calculator. 
# Breaking changes

# Screenshots

zone-aware

order-detail

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:

### DIFF
--- a/packages/vendure-plugin-shipping-extensions/src/services/zone-aware-shipping-tax-calculation.service.ts
+++ b/packages/vendure-plugin-shipping-extensions/src/services/zone-aware-shipping-tax-calculation.service.ts
@@ -1,0 +1,44 @@
+import { Injectable } from '@nestjs/common';
+import {
+  ID,
+  Logger,
+  Order,
+  RequestContext,
+  TaxRate,
+  TransactionalConnection,
+} from '@vendure/core';
+import { loggerCtx } from '../constants';
+
+@Injectable()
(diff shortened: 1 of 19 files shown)


packages/vendure-plugin-shipping-extensions/src/services/zone-aware-shipping-tax-calculation.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  TaxRate,
+  TransactionalConnection,
+} from '@vendure/core';
+import { loggerCtx } from '../constants';
+
+@Injectable()
+export class ZoneAwareShippingTaxCalculationService {
+  constructor(private readonly connection: TransactionalConnection) {}
+
+  async getTaxRateWithCategory(
```
martijnvdbrug, line 16: Shouldn't this `getTaxRateForCategory`? Maybe also add a comment with something like "This resolves the given tax category to a tax rate based on the orders billing Address. If no billing address, it used the shipping address"

martijnvdbrug (approved):
Super nice! Left one comment, plus an additional request:

* Can you also make the distance-based-calculator use a tax-zone instead of a hardcoded tax rate? We also need that one to use the tax category in the CG project

martijnvdbrug (changes requested):
Oops approved, but I meant request Changes

PR #473, opened 2024-07-12
Truncate shipmate payload strings

### LINKED ISSUE #472: Shipmate Plugin: Limit address fields to 40 characters
Fields longer thatn 40 characters throw an error, resulting in an order not being in shipmate. We should truncate the field if it's longer than 40 characters. something like `address?.substring(0,40)` for all fields to Shipmate

dalyathan:
# Description

The changes in this PR implement and close #472 . It will also throw the any shipmate error it might run into so that the job will be  cancelled.

# Breaking changes

# Screenshots

show-failed-job

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


martijnvdbrug (approved):
Perfect, thanks. Can you also update the plugin in the CG project and create a PR?

PR #488, opened 2024-08-01
Invoice Plugin Improvements

### LINKED ISSUE #485: Invoice plugin improvement
We would like the `Regenerate invoice` button to show up as warning/yellow when the order total doesn't match the amount of the latest invoice, so that it's clear to the admin that a new invoice should be generated.

It is not possible to dynamically change the color of a button, but we can create 2 buttons and leverage its [state](https://docs.vendure.io/reference/admin-ui-api/action-bar/action-bar-item#actionbaritem) to hide or show one:
1. Hide the non-styled button when the order-total doesnt match and show the warning button
2. vice versa for the warning button

Image

🔼 Desired when order total doesn't match latest/most-recent invoice total.

Image
🔼 Desired when order total does match, so no action is needed. It should still be possible to click it thougn

dalyathan:
# Description

The changes in this PR implement and close #485
# Breaking changes

# Screenshots

total not match

total-matches

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


packages/vendure-plugin-invoices/src/invoice.plugin.ts
```diff
@@ -84,7 +84,7 @@ export interface InvoicePluginConfig extends InvoicePluginConfigInput {
   },
   compatibility: '>=2.2.0',
   configuration: (config: RuntimeVendureConfig) => {
-    InvoicePlugin.configure(config);
+    void InvoicePlugin.configure(config);
```
martijnvdbrug, line 87: Please us `.catch` here. Using void is dangerous:

When you call an async function, it returns a Promise. If you use void, it effectively discards that Promise.
This means any errors or rejections that occur inside the async function will be unhandled, which could lead to silent failures. Unhandled promise rejections will crash the Node Server

packages/vendure-plugin-invoices/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 3.1.4 (2024-08-01)
+
+- Show `Regenerate invoice` button with warning when order total is greater than the total of the latest invoice ( see #485)
```
martijnvdbrug, line 3: This should be "When order total differs from the latest invoice total" or something similar. It can also be the case that an item was removed from the Order

packages/vendure-plugin-invoices/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@vendure-hub/pinelab-invoice-plugin",
-  "version": "3.1.3",
+  "version": "3.1.4",
```
martijnvdbrug, line 3: Can we make this a new minor? It adds a new feature rather than a bug fix

packages/vendure-plugin-invoices/src/services/invoice.service.ts
```diff
@@ -127,12 +129,13 @@ export class InvoiceService implements OnModuleInit, OnApplicationBootstrap {
    * Listen for OrderPlacedEvents. When an event occures, place generate-invoice job in queue
    */
   onApplicationBootstrap(): void {
-    this.eventBus.ofType(OrderPlacedEvent).subscribe(async ({ ctx, order }) => {
-      this.createInvoiceGenerationJobs(ctx, order.code, 'order-placed');
+    this.eventBus.ofType(OrderPlacedEvent).subscribe(({ ctx, order }) => {
+      void this.createInvoiceGenerationJobs(ctx, order.code, 'order-placed');
```
martijnvdbrug, line 133: Same for this one, should also have a catch.

Good for finding this one, because the previous implementation also had the possibility to crash the node server!

packages/vendure-plugin-invoices/src/services/invoice.service.ts
```diff
@@ -378,7 +389,7 @@ export class InvoiceService implements OnModuleInit, OnApplicationBootstrap {
       );
       if (createCreditInvoiceOnly) {
         // Don't generate normal invoice, so we emit an event now and return
-        this.eventBus.publish(
+        void this.eventBus.publish(
```
martijnvdbrug, line 392: This should be an await

packages/vendure-plugin-invoices/src/services/invoice.service.ts
```diff
@@ -395,7 +406,7 @@ export class InvoiceService implements OnModuleInit, OnApplicationBootstrap {
       order,
       config.templateString!
     );
-    this.eventBus.publish(
+    void this.eventBus.publish(
```
martijnvdbrug, line 409: Also await is best here

packages/vendure-plugin-invoices/src/strategies/load-data-fn.ts
```diff
@@ -71,15 +72,15 @@ export const defaultLoadDataFn: LoadDataFn = async (
   });
   if (!shouldGenerateCreditInvoice) {
     // Normal debit invoice
-    return {
+    return Promise.resolve({
```
martijnvdbrug, line 75: This is probably caused because you removed the `async` above right? This makes it not return a promise anymore.

However, the cleaner way of fixing this is to change the method signature of the LoadDataFn interface to also allow returning of a non promise: `Promise<InvoiceData> | InvoiceData`. This way we don't have to do Promise.resolve for a function that isnt doing anything async

packages/vendure-plugin-invoices/src/strategies/load-data-fn.ts
```diff
@@ -71,15 +72,15 @@ export const defaultLoadDataFn: LoadDataFn = async (
...
       orderDate,
       invoiceNumber: newInvoiceNumber,
       order: order,
-    };
+    });
   }
   // Create credit invoice
   const { previousInvoice, reversedOrderTotals } = shouldGenerateCreditInvoice;
-  return {
+  return Promise.resolve({
```
martijnvdbrug, line 83: Same for this one

packages/vendure-plugin-invoices/src/strategies/local-file-strategy.ts
```diff
@@ -48,11 +47,11 @@ export class LocalFileStrategy implements LocalStorageStrategy {
...
   }
 
-  async streamFile(invoice: InvoiceEntity, res: Response): Promise<ReadStream> {
+  streamFile(invoice: InvoiceEntity, res: Response): Promise<ReadStream> {
     res.set({
       'Content-Type': 'application/pdf',
       'Content-Disposition': `inline; filename="${invoice.invoiceNumber}.pdf"`,
     });
-    return createReadStream(invoice.storageReference);
+    return Promise.resolve(createReadStream(invoice.storageReference));
```
martijnvdbrug, line 55: Same for this, we should change the method signature of streamFile to also allow returning a non-promise ReadStream

packages/vendure-plugin-invoices/src/strategies/s3-storage.strategy.ts
```diff
@@ -15,8 +14,10 @@ export interface Config {
   expiresInSeconds: number;
   bucket: string;
   region?: string;
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
```
martijnvdbrug, line 17: Maybe we should disable some rules like this one that occur a lot. Like the usage of any and unsafe member access for Error logs. We disabled them everywhere, so maybe its better to relax our eslint config a bit.

WDYT?

packages/vendure-plugin-invoices/src/strategies/s3-storage.strategy.ts
```diff
@@ -37,22 +38,27 @@ export class S3StorageStrategy implements RemoteStorageStrategy {
...
   }
 
-  async getPublicUrl(invoice: InvoiceEntity): Promise<string> {
-    return await this.s3!.getSignedUrl('getObject', {
-      Key: invoice.storageReference,
-      Bucket: this.bucket,
-      Expires: this.expiresInSeconds,
-    });
+  getPublicUrl(invoice: InvoiceEntity): Promise<string> {
+    return Promise.resolve(
```
martijnvdbrug, line 55: Also, preferably change method signature in favor of using Promise.resolve

packages/vendure-plugin-invoices/src/ui/util.ts
```diff
@@ -0,0 +1,75 @@
...
+          const showButton = isStyled !== orderTotalMatches;
+          return {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            disabled: order?.state === 'Cancelled',
+            visible: showButton,
+          };
+        })
+      );
+    },
+  });
```
martijnvdbrug, line 74: Nice factory function for adding action bar items!

packages/vendure-plugin-invoices/src/ui/helpers.ts
```diff
@@ -0,0 +1,102 @@
+import gql from 'graphql-tag';
+export const GET_ORDER = gql`
+  query GetOrder($id: ID!) {
+    order(id: $id) {
+      id
```
martijnvdbrug, line 5: Do we really need all these fields? Can't we only fetch invoices?

packages/vendure-plugin-invoices/src/ui/util.ts
```diff
@@ -0,0 +1,75 @@
...
+      );
+      return order$.stream$.pipe(
+        map(({ order }) => {
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+          const latestInvoice = order?.invoices[0];
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+          const orderTotalMatches =
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+            latestInvoice.orderTotals.totalWithTax === order?.totalWithTax;
+          const showButton = isStyled !== orderTotalMatches;
```
martijnvdbrug, line 65: This seems a bit hard to read. Shouldn't it be something like `showButton = isStyled && !orderTotalMatches` (reading as 'show button if isStyled and order total doesn't match').

Also, can we refactor `isStyled` to something like `isWarningButton`? Is styled is a bit ambiguous

packages/vendure-plugin-invoices/src/util/order-calculations.ts
```diff
@@ -9,16 +8,18 @@ import { InvoiceOrderTotals } from '../entities/invoice.entity';
 export function reverseOrderTotals(
   orderTotal: InvoiceOrderTotals
 ): InvoiceOrderTotals {
-  const summaries = orderTotal.taxSummaries.map((summary) => {
+  const summaries = orderTotal.taxSummaries?.map((summary) => {
```
martijnvdbrug, line 11: taxSummaries is always available right? not optional?

packages/vendure-plugin-invoices/src/util/order-calculations.ts
```diff
@@ -9,16 +8,18 @@ import { InvoiceOrderTotals } from '../entities/invoice.entity';
...
     return {
       ...summary,
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
       taxBase: -summary.taxBase,
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
       taxTotal: -summary.taxTotal,
-    };
+    } as InvoiceOrderTotals;
```
martijnvdbrug, line 18: Hmm whats going on here? Why do we need to assert as `InvoiceOrderTotals`? 


Ahh ok, so you are now using the generated type probably? Maybe you can create a Graphql type for the tax summaries also, instead of a scalar, so that we have correct typing for this.

martijnvdbrug (changes requested):
Nice PR! Lot's of changes, lots of improvements. I did leave some more comments. Don't worry, this happens with big PR's

packages/vendure-plugin-invoices/src/ui/helpers.ts
dalyathan: Yes we can

packages/vendure-plugin-invoices/src/util/order-calculations.ts
dalyathan: `OrderTaxSummary` already exists in vendure, so when try to declare it in this plugin, vendure complains.

martijnvdbrug: Ahh yes, I see. Leave it for now, I will check it out locally. 

I think the best thing to do, is to create a new Graphl type `InvoiceTaxSummary`. It happens to be the same as OrderTaxSummary, but they have no relation to eachother, so they should actually be different types. I was lazy, so I originally just reused the OrderTaxSummary :upside_down_face: 

That should solve it. We can then also get rid of the InvoiceOrderTotals in the invoice entity, and only use the generated type

martijnvdbrug: @dalyathan I think this fixed most of the issues: https://github.com/Pinelab-studio/pinelab-vendure-plugins/pull/488/commits/134c80ba45d635f8519ecd1ddb5454b0f49d558b

packages/vendure-plugin-invoices/src/strategies/storage-strategy.ts
```diff
@@ -40,7 +40,7 @@ export interface RemoteStorageStrategy extends BaseStorageStrategy {
    * For example an downloadLink to a Google storage bucket
    * or Amazon S3 instance
    */
-  getPublicUrl(invoice: InvoiceEntity): Promise<string>;
+  getPublicUrl(invoice: InvoiceEntity): Promise<string> | string;
```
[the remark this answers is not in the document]
martijnvdbrug: Bad advice. The calling code can just always await. This return type gives the developer the flexibility to decide if he needs async or sync strategy

PR #510, opened 2024-09-18
Optionally including plugins 

### LINKED ISSUE #452: Acceptblue plugin and stripe plugin make it very hard to be optionally included
**Which plugin/repository is the issue about?**
accept-blue and stripe

**Describe the bug**
When you have various payment services that you are using, in order to get a PaymentMethodQuote, you need to use the 
```
stripeSubscriptionPublishableKey
acceptBlueHostedTokenizationKey
```

However, when you use accept blue in some channels and not in others, you won't have an accept blue in your list of payment methods and this line will throw a hard "missing accept blue config" error (presumably the same for stripe).
https://github.com/Pinelab-studio/pinelab-vendure-plugins/blob/600dce99a469d1b521496f04558f1e26cbdb0174/packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts#L352

**To Reproduce**

1. Create a channel
2. Add a dummy payment handler
3. Do not add an accept blue payment handler
4. Add a client side graphql something like this:

```
fragment paymentMethodQuoteFragment on PaymentMethodQuote {
    id
    name
    code
    isEligible
    eligibilityMessage
    stripeSubscriptionPublishableKey
    acceptBlueHostedTokenizationKey
    customFields
  }
```

**Expected behavior**
What you don't want is a hard error when you are retrieving the `acceptBlueHostedTokenizationKey`, because that is hard coded and there is no `AcceptBlueMethodQuote` to filter by. 

One suggestion might be to only throw the missing AcceptBlue error in case you're actually trying to call it for a payment action instead of reading static data.

Another suggestion might be to return an ErrorResult with a `missing accept blue configuration` style error

**Environment**

- NodeJS version
- Vendure version: 2.2.6
- Plugin version 1.3.1
- Database postgres
- Any other version numbers from the environment you're using

dalyathan:
# Description

The changes in this PR implement and close #452 
# Breaking changes

# Screenshots

# Checklist

📌 Always:

👍 Most of the time:

📦 For publishable packages:


packages/vendure-plugin-accept-blue/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.6.2 (2024-09-18)
+
+- Fix #452
```
martijnvdbrug, line 3: Can you make this a bit more descriptive, so that its also clear what is fixed without reading the issue?

packages/vendure-plugin-stripe-subscription/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.5.3 (2024-09-18)
+
+- Fix #452
```
martijnvdbrug, line 3: Same for this. Bit more descriptive please :)

martijnvdbrug (changes requested):
Nice! Minor improvements requested

martijnvdbrug (approved):
Thanks again!

PR #541, opened 2024-12-04
feat: add support for discount code on Klaviyo's order placed event

nicohps:
Klaviyo's order placed event supports a DiscountCode. This PR adds support for this.

### DIFF
--- a/packages/vendure-plugin-klaviyo/CHANGELOG.md
+++ b/packages/vendure-plugin-klaviyo/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.2.1 (2024-12-04)
+
+- Add support for DiscountCode on Placed Order event
+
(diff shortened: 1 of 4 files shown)


martijnvdbrug (changes requested):
Can you increase the version number in package.json + add a line in the Changelog file? This way it gets auto-published

packages/vendure-plugin-klaviyo/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 1.2.0 (2024-12-04)
```
martijnvdbrug, line 1: Version isn't same as package :-) Should be 1.2.1

martijnvdbrug (approved):
Thanks!

PR #559, opened 2025-01-08
feat(accept-blue): add ctx for event, log order history entry

mschipperheyn:
# Description

We want to better understand what happened if a payment doesn't go through for some reason and be able to view the history of this. Therefore, I added a logHistoryEntry for accept blue
Also, I added ctx to the event. This would allow us to capture the event and affect channel specific processing.

# To do before merge

I didn't write tests for this (yet). Please have a look and see if you think we can proceed with this change and I can write a test

No update has been done yet to README or version

# Breaking changes

# Screenshots

# Checklist

📌 Always:
- [ x] Set a clear title
- [x ] I have checked my own PR

👍 Most of the time:
- [ ] Updated the README

📦 For publishable packages:

### DIFF
--- a/packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
+++ b/packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
@@ -11,7 +11,6 @@ import {
   ID,
   Logger,
   Order,
-  OrderHistoryEntryData,
   OrderLine,
   PaymentMethod,
   PaymentMethodEvent,
@@ -30,7 +29,6 @@ import { SubscriptionHelper } from '../';
 import { AcceptBluePluginOptions } from '../accept-blue-plugin';
 import { PLUGIN_INIT_OPTIONS, loggerCtx } from '../constants';
 import {
-  ACCEPT_BLUE_TRANSACTION_UPDATE,
   AcceptBlueChargeTransaction,
   AcceptBlueEvent,
   AcceptBluePaymentMethod,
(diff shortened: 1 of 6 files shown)


packages/vendure-plugin-accept-blue/src/api/api-extensions.ts
```diff
@@ -118,6 +118,10 @@ export const commonApiExtensions = gql`
     savedAcceptBluePaymentMethods: [AcceptBluePaymentMethod!]!
   }
 
+  extend enum HistoryEntryType {
+    ACCEPT_BLUE_TRANSACTION_UPDATE
```
martijnvdbrug, line 122: We need a custom Admin UI component for this, or otherwise it throws an error in Admin Ui right? Or am I mistaken?

If we do, it might be easier to just add it as order note, or mis-use any of the other built in entries.

mschipperheyn: Ah yeah, The f**** angular crap. I don't work with that.

packages/vendure-plugin-accept-blue/src/api/accept-blue-service.ts
```diff
@@ -62,6 +63,7 @@ export class AcceptBlueService implements OnApplicationBootstrap {
     private readonly customerService: CustomerService,
     private readonly entityHydrator: EntityHydrator,
     private connection: TransactionalConnection,
+    private historyService: HistoryService,
```
martijnvdbrug, line 66: Can you remove this, and the import, as this change doesnt require historyService anymore?

martijnvdbrug (approved):
Perfect!

PR #594, opened 2025-04-17
feat(stripe-subscription): avoid webhook secret validation

martijnvdbrug:
# 2.8.0 (2025-04-17)

- Don't validate webhook secret, but call Stripe API directly to get valid data.
- Skip job queue for subscription creation, but create synchronously instead, for direct feedback.

# Checklist

📌 Always:
- [ ] Set a clear title

👍 Most of the time:
- [ ] Updated the README

📦 For publishable packages:


kamui:
I didn't have a chance to dig too much into this, but I upgraded to 2.8.0 with this change and noticed several errors where no orders were placed during the time I deployed it and this is the error I got from Sentry:

image

PR #595, opened 2025-04-18
docs: add troubleshooting section for Admin UI 403 error with bearer …

Guts996:
# Description

This PR updates the documentation for the Order PDF's Plugin to include a troubleshooting note for Admin UI 403 errors related to authentication.

Specifically, it explains that if a user encounters `403 Forbidden` errors in the Admin UI despite having the correct permissions, they likely need to explicitly set the `tokenMethod: 'bearer'` option in `adminUiConfig`. This resolves the issue when using bearer tokens instead of cookies for authentication.

# To do before merge

- [ ] Confirm formatting and rendering in the live docs environment

# Breaking changes

None

# Screenshots

_Not applicable – docs only_

# Checklist

📌 Always:
- [x] Set a clear title  
- [x] I have checked my own PR

👍 Most of the time:
- [x] Updated the README

📦 For publishable packages:

### DIFF
--- a/packages/vendure-plugin-order-pdfs/README.md
+++ b/packages/vendure-plugin-order-pdfs/README.md
@@ -39,6 +39,27 @@ plugins: [
 
 You can also download multiple PDF's in the order list. This is currently limited to 10 files, because PDF generation runs in the main instance, not the worker.
 
+## Troubleshooting
+
+### 403 Error Despite Proper Permissions
+
+If you're encountering a `403 Forbidden` error in the Admin UI even after granting the correct permissions to your user (e.g., SuperAdmin), you may need to explicitly configure the token method used by the Admin UI to match your server configuration.
+
+Add the following to your `AdminUiPlugin` configuration to ensure compatibility:
+
+```ts
+AdminUiPlugin.init({
+  adminUiConfig: {
+    tokenMethod: 'bearer', // ← switch from default 'cookie' to match server settings
+    // The following are the defaults and should match your server config
+    authTokenHeaderKey: 'vendure-auth-token',
+    channelTokenKey: 'vendure-token',
+  },
+}),
+```
+
+This ensures that the Admin UI and the Vendure server are using the same token method (`bearer`), which is particularly important if your server is configured with both `"bearer"` and `"cookie"` token methods.
+
 ## Docker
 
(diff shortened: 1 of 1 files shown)


packages/vendure-plugin-order-pdfs/README.md
```diff
@@ -39,6 +39,27 @@ plugins: [
...
+### 403 Error Despite Proper Permissions
+
+If you're encountering a `403 Forbidden` error in the Admin UI even after granting the correct permissions to your user (e.g., SuperAdmin), you may need to explicitly configure the token method used by the Admin UI to match your server configuration.
+
+Add the following to your `AdminUiPlugin` configuration to ensure compatibility:
+
+```ts
+AdminUiPlugin.init({
+  adminUiConfig: {
+    tokenMethod: 'bearer', // ← switch from default 'cookie' to match server settings
```
martijnvdbrug, line 53: Does `tokenMethod: ['bearer'. 'cookie']` also work? If so, can we update this section to describe that `bearer` is **also** needed for this plugin. This way we don't have to force people to migrate to bearer token, but they can just enable bearer next to cookies.

martijnvdbrug (changes requested):
@Guts996 Thank you so much for making an effort on improving the open source ecosystem. Much appreciated!

I have added 1 comment/requested change. Can you perhaps take a look at that?

packages/vendure-plugin-order-pdfs/README.md
martijnvdbrug: @Guts996 I would believe you, if I hadn't coincidentally used this yesterday in one of our plugins :upside_down_face: 

It actually is possible since V1.2.0: https://docs.vendure.io/reference/typescript-api/auth/auth-options
> From v1.2.0 it is possible to specify both methods as a tuple: ['cookie', 'bearer'].

Guts996: well, I'm new to vendure, so please correct me if I'm wrong: 
what you shared is a part of the plugin options for the backend and indeed it's possible have both cookie and bearer but I'm talking about the admin ui plugin. 
see this: https://docs.vendure.io/reference/typescript-api/common/admin-ui/admin-ui-config#tokenmethod

martijnvdbrug: I am terribly sorry, you are right!

martijnvdbrug (commented):
Approved

PR #648, opened 2025-11-07
QLS Fulfillment plugin

martijnvdbrug:
:warning: Awaiting real life testing in client project before merge + e2e tests

# QLS Fulfillment plugin

## Product Sync

* Full product sync (via Admin UI):
  * Fetches all products from QLS
  * Ensures corresponding Vendure variants exist
  * Updates Vendure stock levels from QLS
* Partial product sync: Creates/updates products in QLS when products change in Vendure
* Partial stock sync: Updates Vendure stock based on incoming QLS webhooks

## Order Sync

* Pushes orders to QLS automatically on order placement
* Supports manual order push to QLS via Admin UI
* Updates order status in Vendure based on QLS webhooks

# Checklist

📌 Always:
- [ ] Set a clear title

👍 Most of the time:
- [ ] Updated the README

📦 For publishable packages:

### DIFF
--- a/packages/vendure-plugin-qls-fulfillment/src/api/qls-webhooks-controller.ts
+++ b/packages/vendure-plugin-qls-fulfillment/src/api/qls-webhooks-controller.ts
@@ -7,7 +7,12 @@ import {
   Query,
   Req,
 } from '@nestjs/common';
-import { ChannelService, Logger, RequestContext } from '@vendure/core';
+import {
+  ChannelService,
+  ForbiddenError,
+  Logger,
+  RequestContext,
+} from '@vendure/core';
 import { asError } from 'catch-unknown';
 import { Request } from 'express';
 import { QlsProductService } from '../services/qls-product.service';
@@ -39,10 +44,11 @@ export class QlsWebhooksController {
     @Body() body: IncomingStockWebhook | IncomingOrderWebhook
   ) {
     if (webhookSecret !== this.options.webhookSecret) {
-      return Logger.warn(
+      Logger.warn(
         `Incoming webhook with invalid secret for channel '${channelToken}' to '${request.url}'`,
         loggerCtx
       );
+      throw new ForbiddenError();
     }
     try {
       const ctx = await this.getCtxForChannel(channelToken);
--- a/packages/vendure-plugin-qls-fulfillment/src/services/qls-order.service.ts
+++ b/packages/vendure-plugin-qls-fulfillment/src/services/qls-order.service.ts
@@ -105,6 +105,10 @@ export class QlsOrderService implements OnModuleInit, OnApplicationBootstrap {
     }
   }
 
+  /**
+   * Push an order to QLS by id.
+   * Returns a human-readable message describing the result of the operation (Used as job result).
+   */
   async pushOrderToQls(ctx: RequestContext, orderId: ID): Promise<string> {
     const client = await getQlsClient(ctx, this.options);
     if (!client) {
@@ -152,6 +156,11 @@ export class QlsOrderService implements OnModuleInit, OnApplicationBootstrap {
           });
         })
       );
+      if (qlsProducts.length === 0) {
+        const message = `No products to push to QLS for order '${order.code}'. Ignoring order.`;
+        Logger.info(message, loggerCtx);
+        return message;
+      }
       const additionalOrderFields =
         await this.options.getAdditionalOrderFields?.(
           ctx,
--- a/packages/vendure-plugin-qls-fulfillment/package.json
+++ b/packages/vendure-plugin-qls-fulfillment/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pinelab/vendure-plugin-qls-fulfillment",
-  "version": "1.0.0-beta.23",
+  "version": "1.0.0",
   "description": "Vendure plugin to fulfill orders via QLS.",
   "keywords": [
     "fulfillment",
--- a/packages/vendure-plugin-qls-fulfillment/test/e2e.spec.ts
+++ b/packages/vendure-plugin-qls-fulfillment/test/e2e.spec.ts
@@ -1,16 +1,18 @@
 import { DefaultLogger, LogLevel, mergeConfig } from '@vendure/core';
 import {
   createTestEnvironment,
+  E2E_DEFAULT_CHANNEL_TOKEN,
   registerInitializer,
   SimpleGraphQLClient,
   SqljsInitializer,
   testConfig,
   TestServer,
 } from '@vendure/testing';
-import { beforeAll, expect, it } from 'vitest';
+import { afterEach, beforeAll, expect, it } from 'vitest';
 import { initialData } from '../../test/src/initial-data';
-import { QlsPlugin } from '../src';
+import { FulfillmentProduct, QlsPlugin } from '../src';
 import { testPaymentMethod } from '../../test/src/test-payment-method';
+import { gql } from 'graphql-tag';
 
 let server: TestServer;
 let adminClient: SimpleGraphQLClient;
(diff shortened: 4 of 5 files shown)


packages/vendure-plugin-qls-fulfillment/src/custom-fields.ts
```diff
@@ -0,0 +1,23 @@
...
+  {
+    name: 'qlsProductId',
+    type: 'string',
+    label: [{ value: 'QLS Product ID', languageCode: LanguageCode.en }],
+    nullable: true,
+    public: false,
+    readonly: true,
+    ui: { tab: 'QLS' },
+  },
+];
```
martijnvdbrug, line 33: @alexanderkraus this was re-introduced:

* Rate limit for getting products is 60/minute.
* An order with 10 items thus requires 10 calls to the `get fulfillment products` endpoint
* This means we can only have 6 orders with 10 items placed in the same minute. This doesn't happen often, but changes are very high that it will occur at some point, especially in combination with the global rate limit of 500/5 minutes.

This custom field, as you introduced it, allows us to get the productId from our own database when creating orders. This prevents the 10 calls per order described above.

packages/vendure-plugin-qls-fulfillment/src/api/qls-admin.resolver.ts
```diff
@@ -0,0 +1,32 @@
...
+  @Transaction()
+  @Allow(fullSyncPermission.Permission)
+  async triggerQlsProductSync(@Ctx() ctx: RequestContext) {
+    await this.qlsService.triggerFullSync(ctx);
+    return true;
+  }
+
+  @Mutation()
+  @Transaction()
+  @Allow(fullSyncPermission.Permission)
```
alexanderkraus: Should there be a separate permission for pushing orders to QLS?

packages/vendure-plugin-qls-fulfillment/src/api/qls-webhooks-controller.ts
```diff
@@ -0,0 +1,100 @@
...
+export class QlsWebhooksController {
+  constructor(
+    private channelService: ChannelService,
+    private qlsProductService: QlsProductService,
+    private qlsOrderService: QlsOrderService,
+    @Inject(PLUGIN_INIT_OPTIONS) private options: QlsPluginOptions
+  ) {}
+
+  /**
+   * Endpoint for all incoming webhooks
```
alexanderkraus, line 37: If we had distinct methods for different types of webhooks, we would not necessarily need to inspect the request body to distinguish between webhook events.

We could still have a "catchall" method for unknown webhook events, but the user would anyways need to select the type and URL of each webhook individually in the QLS admin panel.

packages/vendure-plugin-qls-fulfillment/src/custom-fields.ts
```diff
@@ -0,0 +1,23 @@
+import { CustomFieldConfig, LanguageCode } from '@vendure/core';
+import {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  CustomProductVariantFields,
+} from '@vendure/core/dist/entity/custom-entity-fields';
```
alexanderkraus, line 9: What is this type import needed for?

packages/vendure-plugin-qls-fulfillment/src/lib/qls-client.ts
```diff
@@ -0,0 +1,156 @@
...
+      const result = await this.rawRequest<FulfillmentProduct[]>(
+        'GET',
+        `fulfillment/products?page=${page}`
+      );
+      if (!result.data || result.data.length === 0) {
+        break;
+      }
+      allProducts.push(...result.data);
+      hasNextPage = result.pagination?.nextPage ?? false;
+      page++;
```
alexanderkraus, line 77: Depending on the speed of requests, we might need to introduce a delay here to mitigate the risk of running into a rate limit.

packages/vendure-plugin-qls-fulfillment/src/qls-plugin.ts
```diff
@@ -0,0 +1,60 @@
...
+    resolvers: [QlsAdminResolver],
+  },
+})
+export class QlsPlugin {
+  static options: QlsPluginOptions;
+
+  constructor(
+    private eventBus: EventBus,
+    private qlsService: QlsProductService
+  ) {}
```
alexanderkraus: The injected dependencies are not used.

I personally prefer registering event listeners in the plugin (instead of in the services) and let services only perform the actual work to be done. This would allow having everything event related in one place and (if wanted) use a shared job queue for all jobs, which are related to this plugin.

But I guess, that's a matter o taste 🙂

packages/vendure-plugin-qls-fulfillment/src/services/qls-order.service.ts
```diff
@@ -0,0 +1,245 @@
...
+          name: name,
+          companyname: order.shippingAddress.company ?? '',
+          street: order.shippingAddress.streetLine1 ?? '',
+          housenumber: order.shippingAddress.streetLine2 ?? '',
+          postalcode: order.shippingAddress.postalCode ?? '',
+          locality: order.shippingAddress.city ?? '',
+          country: order.shippingAddress.countryCode?.toUpperCase() ?? '',
+          email: order.customer?.emailAddress ?? '',
+          phone: order.customer?.phoneNumber ?? '',
+        },
```
alexanderkraus: Sending empty fields to QLS will probably work technically, but won't this increase the risk of hard t o debug issues during the actual fulfillment process? Maybe something worth to mention in the README?

packages/vendure-plugin-qls-fulfillment/src/api/qls-admin.resolver.ts
martijnvdbrug: I think that makes sense, yes. Let me add that :mechanical_arm:

packages/vendure-plugin-qls-fulfillment/src/api/qls-webhooks-controller.ts
martijnvdbrug: Valid points, I've though about this: I chose having a single endpoint, because it is less prone to errors when users copy/paste webhooks into the QLS UI. 

I favored user simplicity over code simplicity. I guess there is no right or wrong here, but I prefer to keep it this way :upside_down_face:

packages/vendure-plugin-qls-fulfillment/src/custom-fields.ts
martijnvdbrug: To be honest: I am not exactly sure how this works, but [the Vendure Docs](https://docs.vendure.io/guides/developer-guide/custom-fields/#typescript-typings) say this is needed for proper typescript support:

```ts

// Note: we are using a deep import here, rather than importing from `@vendure/core` due to
// a possible bug in TypeScript (https://github.com/microsoft/TypeScript/issues/46617) which
// causes issues when multiple plugins extend the same custom fields interface.
```

I've added this comment to our code as well.

packages/vendure-plugin-qls-fulfillment/src/lib/qls-client.ts
martijnvdbrug: Yes, good one!

packages/vendure-plugin-qls-fulfillment/src/qls-plugin.ts
martijnvdbrug: I removed the unused dependencies.

Valid points, I don't really have an opinion on that. Currently, the QLS order service listens for events that relate to orders, and QLS product service what relates to products. The path of least resistance is to leave it like that now :laughing:

packages/vendure-plugin-qls-fulfillment/src/services/qls-order.service.ts
martijnvdbrug: Good points. I have added validation for required fields!

PR #652, opened 2026-01-02
fix: prefix delivery options with tag key

nicohps:
Fixes the delivery options. Not consistent with the QLS Redoc but it was communicated to implement it like this by QLS IT support


martijnvdbrug (changes requested):
Approved! Can you update the package.json version and changelog entry? Then it will be published

PR #661, opened 2026-01-23
Store Credit Plugin Setup

dalyathan:
# Description

* Store credit in wallets conneted to customers

# Checklist

📌 Always:
- [X] Set a clear title
- [X] I have checked my own PR

👍 Most of the time:
- [ ] Updated the README

📦 For publishable packages:
- [X] Increased the version number in `package.json`


packages/vendure-plugin-store-credit/src/api/api-extensions.ts
```diff
@@ -0,0 +1,55 @@
...
+export const scalars = gql`
+  scalar DateTime
+  scalar Money
+  scalar CurrencyCode
+`;
+
+export const commonApiExtension = gql`
+  type Wallet {
+    id: ID!
+    createdAt: DateTime!
```
martijnvdbrug, line 15: Can we add a `name` to the wallet? We want to be able to display wallets to admins and customers, so they need some way to identify their wallets.

packages/vendure-plugin-store-credit/src/api/api-extensions.ts
```diff
@@ -0,0 +1,55 @@
...
+export const adminApiExtensions = gql`
+  ${commonApiExtension}
+
+  type WalletAdjustment {
+    id: ID!
+    createdAt: DateTime!
+    updatedAt: DateTime!
+    wallet: Wallet!
+    amount: String!
+  }
```
martijnvdbrug, line 53: `WalletAdjustment` can be in common api extensions: Customers will also be allowed to view mutation to **their own** wallets

packages/vendure-plugin-store-credit/src/api/api-extensions.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  input CreateWalletInput {
+    customerId: ID!
+    initialBalance: Money!
```
martijnvdbrug: Initial balance is always 0. From there on we can mutate the balance, so that also the initial balance shows up as mutation.

And, can we accept a `name` as input also?

packages/vendure-plugin-store-credit/src/api/api-extensions.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const commonApiExtension = gql`
+  type Wallet {
+    id: ID!
+    createdAt: DateTime!
+    updatedAt: DateTime!
+    customer: Customer!
+    currencyCode: CurrencyCode!
+    balance: Money!
+  }
```
martijnvdbrug, line 37: A wallet should have `adjustments: [WalletAdjustment!]!`. But similar to Vendure's variant, this subfield should accept list pagination. When no pagination is given, we only return the most recent 10 adjustments.

packages/vendure-plugin-store-credit/src/api/customer-entity.resolver.ts
```diff
@@ -0,0 +1,24 @@
...
+import {
+  Ctx,
+  RequestContext,
+  Customer,
+  TransactionalConnection,
+} from '@vendure/core';
+import { Wallet } from '../entities/wallet.entity';
+
+@Resolver('Customer')
+export class CustomerEntityResolver {
```
martijnvdbrug: Can we seperate resolvers into `common-resolvers.ts` and `admin-resolvers.ts`? This customer field resolver can then be inlined in `common-resolver`:

```ts
  @ResolveField()
  @Resolver('Customer')
  async wallets(
```

packages/vendure-plugin-store-credit/src/api/customer-entity.resolver.ts
```diff
@@ -0,0 +1,24 @@
...
+  @ResolveField()
+  wallets(
+    @Ctx() ctx: RequestContext,
+    @Parent() customer: Customer
+  ): Promise<Wallet[]> {
+    return this.connection.getRepository(ctx, Wallet).find({
+      where: { customer: { id: customer.id } },
+      relations: ['ledgerEntries'],
+    });
+  }
```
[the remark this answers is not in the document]
martijnvdbrug: No need: When a someone can fetch the customer, he can also view wallet balance. Only admins can view customers, and a logged in customer can view his own profile.

packages/vendure-plugin-store-credit/src/api/wallet.resolver.ts
```diff
@@ -0,0 +1,27 @@
...
+    return this.walletService.createWallet(ctx, args);
+  }
+
+  @Transaction()
+  @Mutation()
+  adjustBalanceForWallet(
+    @Ctx() ctx: RequestContext,
+    @Args() args: AdjustBalanceForWalletInput
+  ) {
+    return this.walletService.adjustBalanceForWallet(ctx, args);
```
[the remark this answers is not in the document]
martijnvdbrug: @dalyathan Good point, this should be `@Args('input')`

packages/vendure-plugin-store-credit/src/entities/wallet.entity.ts
```diff
@@ -0,0 +1,54 @@
...
+  channel!: Channel;
+
+  @Column({ type: 'enum', enum: CurrencyCode })
+  currencyCode!: CurrencyCode;
+
+  @Column({ type: 'bigint', default: 0 })
+  balance!: string;
+
+  @OneToMany(() => WalletAdjustment, (e) => e.wallet)
+  ledgerEntries!: WalletAdjustment[];
```
martijnvdbrug: Lets name this `adjustments` to keep it consistent across GraphQL and the rest of the codebase.

_(It is not wrong to call it a `ledger`, but since we have named it adjustment in Grapqhl, we might as well do it here too)_

packages/vendure-plugin-store-credit/src/entities/wallet.entity.ts
```diff
@@ -0,0 +1,54 @@
...
+  @JoinTable()
+  channels!: Channel[];
+
+  @ManyToOne(() => Customer, { onDelete: 'CASCADE', nullable: false })
+  @Index()
+  customer!: Customer;
+
+  @ManyToOne(() => Channel, { onDelete: 'CASCADE', nullable: false })
+  @Index()
+  channel!: Channel;
```
martijnvdbrug: Good point, hadn't thought about this. Can you add an e2e testcase that we test this? (can be a placeholder, just so we dont forget):

Testcases: 
* Assigning a customer to another channel should not show the customers wallets in the other channel
* A customer should **not** be able to spend store credit of Channel A in Channel B << This one seems like an important security check.

packages/vendure-plugin-store-credit/src/services/wallet.service.ts
```diff
@@ -0,0 +1,82 @@
...
+  async createWallet(
+    ctx: RequestContext,
+    input: CreateWalletInput
+  ): Promise<Wallet> {
+    const customer = await this.connection
+      .getRepository(ctx, Customer)
+      .findOneOrFail({ where: { id: input.customerId } });
+    const wallet = new Wallet({
+      customer,
+      initialAmount: input.initialBalance as number,
```
martijnvdbrug: Lets make this always 0

packages/vendure-plugin-store-credit/src/services/wallet.service.ts
```diff
@@ -0,0 +1,82 @@
...
+      amount: input.initialBalance as number,
+    });
+    return this.connection.getRepository(ctx, Wallet).save(wallet);
```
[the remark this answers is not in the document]
martijnvdbrug: We will see when we start testing!

packages/vendure-plugin-store-credit/src/services/wallet.service.ts
```diff
@@ -0,0 +1,82 @@
...
+    const adjustmentRepo = this.connection.getRepository(ctx, WalletAdjustment);
+
+    if (input.adjustmentType === 'CREDIT') {
+      const res = await walletRepo
+        .createQueryBuilder()
+        .update(Wallet)
+        .set({ balance: () => `balance + :amount` })
+        .where('id = :id', { id: input.walletId })
+        .setParameters({ amount: input.amount as number })
+        .execute();
```
martijnvdbrug: Do we know if this works for all DB's (sqlite, postgres, mysql)? If not, ignore for now, I can test it locally.

packages/vendure-plugin-store-credit/src/api/api-extensions.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  enum WalletAdjustmentType {
+    CREDIT
+    DEBIT
+  }
+
+  input AdjustBalanceForWalletInput {
+    amount: Money!
+    adjustmentType: WalletAdjustmentType!
```
martijnvdbrug: In most stock systems, a user can just do a negative number, so the input can be `10` of `-10`. Lets do that here as well.


We can remove the enum CREDIT/DEBIT. While correct terms in bookkeeping, I doubt if all Vendure admins users will find this intuitive

packages/vendure-plugin-store-credit/src/services/wallet.service.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      if ((res.affected ?? 0) !== 1) {
+        throw new UserInputError('Wallet not found');
+      }
+    }
+
+    if (input.adjustmentType === 'DEBIT') {
```
martijnvdbrug: We don't need DEBIT/CREDIT checks anymore if we allow negative input I believe?

packages/vendure-plugin-store-credit/src/services/wallet.service.ts
```diff
@@ -0,0 +1,82 @@
...
+        .createQueryBuilder()
+        .update(Wallet)
+        .set({ balance: () => `balance + :amount` })
+        .where('id = :id', { id: input.walletId })
+        .setParameters({ amount: input.amount as number })
+        .execute();
+
+      if ((res.affected ?? 0) !== 1) {
+        throw new UserInputError('Wallet not found');
+      }
```
martijnvdbrug: Can we explicitly check for existence of a wallet beforehand? Invalid input.amount (or other errors) might also cause affected rows to be 0, which makes this error throwing incorrect.

packages/vendure-plugin-store-credit/src/services/wallet.service.ts
```diff
@@ -0,0 +1,82 @@
...
+        .update(Wallet)
+        .set({ balance: () => `balance - :amount` })
+        .where('id = :id', { id: input.walletId })
+        .andWhere('balance >= :amount')
+        .setParameters({ amount: input.amount as number })
+        .execute();
+
+      if ((res.affected ?? 0) !== 1) {
+        throw new UserInputError('Insufficient wallet balance');
+      }
```
martijnvdbrug: Maybe we can add a log statement here, logging the `res`? It might contain other information about errors. Perhaps not all of them will be caused by 'insufficient balance', not sure.

packages/vendure-plugin-store-credit/src/services/wallet.service.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    await adjustmentRepo.save(
+      adjustmentRepo.create({
+        wallet: { id: input.walletId } as Wallet,
+        amount: input.amount as number,
+      })
```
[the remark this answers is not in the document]
martijnvdbrug: Not sure, we will test this later

martijnvdbrug, line 150: This needs to be in the same transaction as `.set({ balance: () => `balance - :amount` })`, but that DB statement currently doesn't use `ctx`, so I believe it will not be rolled back when this adjustment fails?

Please add an e2e (or placeholder for now) as well: If adjustment saving fails, balance increment/decrement should be rolled back

packages/vendure-plugin-store-credit/src/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const BETTER_SEARCH_PLUGIN_OPTIONS = Symbol(
+  'BETTER_SEARCH_PLUGIN_OPTIONS'
+);
+export const loggerCtx = 'BetterSearchPlugin';
```
martijnvdbrug: Please rename these to StoreCreditPlugin :)

packages/vendure-plugin-store-credit/src/services/wallet.service.ts
dalyathan: so remove the `initialAmount` field  and rename `amount` field to `balance`?

packages/vendure-plugin-store-credit/src/services/wallet.service.ts
dalyathan: [`WalletResolver.adjustBalanceForWallet`](https://github.com/Pinelab-studio/pinelab-vendure-plugins/blob/9fff672254cf6c5b7479ca0fea6bf9e30a6d0147/packages/vendure-plugin-store-credit/src/api/wallet.resolver.ts#L19) is wrapped with ` @Transaction()`, wont that cut it?

martijnvdbrug: That is a good question. Honestly, I don't know. I guess we have to test it out.

PR #693, opened 2026-05-29
fix(stock-monitoring): fix widget sorting

LazyClicks:
# Description
Fixes stock monitoring sorting so that out-of-stock items ordered from lowest to highest stock

# To do before merge

# Breaking changes

# Screenshots

# Checklist

📌 Always:
- [x ] Set a clear title
- [ x] I have checked my own PR

👍 Most of the time:
- [ ] Updated the README

📦 For publishable packages:
- [x ] Increased the version number in `package.json`
- [x ] Added changes to the `CHANGELOG.md`


packages/vendure-plugin-stock-monitoring/src/services/stock-monitoring.service.ts
```diff
@@ -124,6 +124,15 @@ export class StockMonitoringService
...
     const variants = await this.variantService.findByIds(ctx, variantIds);
+    // findByIds does not preserve input order, so re-sort to match the cached ASC order
+    const idOrder = new Map(
+      variantIds.map((id, index) => [id.toString(), index])
+    );
+    variants.sort(
+      (a, b) =>
+        (idOrder.get(a.id.toString()) ?? 0) -
+        (idOrder.get(b.id.toString()) ?? 0)
+    );
```
martijnvdbrug, line 135: Good catch. So `variantIds` has the correct sort order, but doing `findByIds` messes it up?

martijnvdbrug (changes requested):
@LazyClicks Can you revert all changes except the stock monitoring ones? There is a lot of unrelated formatting changes here.

Something like `git reset --hard origin/main <filepath>` should exists. Not sure what the exact command is.

packages/vendure-plugin-stock-monitoring/src/services/stock-monitoring.service.ts
LazyClicks: > Good catch. So `variantIds` has the correct sort order, but doing `findByIds` messes it up?

yeah findbyids does not preserve the input order as its output

PR #694, opened 2026-06-03
feat(invoice-plugin): Allow overriding Chrome executable path for Pup…

LazyClicks:
…peteer

# Description

# To do before merge

# Breaking changes

# Screenshots

# Checklist

📌 Always:
- [ ] Set a clear title

👍 Most of the time:
- [ ] Updated the README

📦 For publishable packages:


martijnvdbrug (changes requested):
Nice work on the extra puppeteer config. Can you bump the version to next major? (Also changelog). I dont think we should release this as patch, as it requires consumers to update their docker file

martijnvdbrug (approved):

PR #695, opened 2026-06-09
fix(invoice-plugin): Fix missing invoice files causes server crashing + Duplicate invoice generation on concurrent orders

LazyClicks:
# Description

# To do before merge

# Breaking changes

# Screenshots

# Checklist

📌 Always:
- [ ] Set a clear title

👍 Most of the time:
- [ ] Updated the README

📦 For publishable packages:

### DIFF
--- a/packages/vendure-plugin-invoices/src/services/invoice.service.ts
+++ b/packages/vendure-plugin-invoices/src/services/invoice.service.ts
@@ -473,6 +473,7 @@ export class InvoiceService implements OnModuleInit, OnApplicationBootstrap {
   /**
    * Create an invoice and save it's reference on an order in the database.
    * Passing a `previousInvoice` will generate a credit invoice.
+   * Uses a retry loop to handle duplicate invoice number conflicts from concurrent inserts.
    */
   private async createAndSaveInvoice(
     ctx: RequestContext,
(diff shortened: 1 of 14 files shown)


packages/vendure-plugin-invoices/src/services/invoice.service.ts
```diff
@@ -493,43 +494,72 @@ export class InvoiceService implements OnModuleInit, OnApplicationBootstrap {
...
+        // Save storage reference on the invoice row
+        const invoiceRepo = this.connection.getRepository(ctx, InvoiceEntity);
+        await invoiceRepo.update(invoiceRowId, { storageReference });
+        Logger.info(
+          `Created ${
+            isCreditInvoiceFor ? 'credit ' : ' '
+          }invoice ${invoiceNumber} for order ${order.code}`,
+          loggerCtx
+        );
+        return await invoiceRepo.findOneOrFail({
```
martijnvdbrug, line 534: @LazyClicks I think we also need a safeRemove here? We are only removing it when an error is caught, but we can also remove the tempfile after saving the storage reference I guess? The file is already saved to its 'real' location with `this.config.storageStrategy.save`

packages/vendure-plugin-invoices/test/e2e.spec.ts
```diff
@@ -448,6 +448,24 @@ describe('Download invoices', function () {
...
+      where: { invoiceNumber: 10001 },
+    });
+    expect(invoice).toBeDefined();
+    // Delete the invoice file from disk
+    await fs.unlink(invoice!.storageReference);
+    // Attempt to download - should return 403 instead of crashing the server
+    const res = await fetch(
+      `http://localhost:3106/invoices/e2e-default-channel/${order.code}/10001?email=hayden.zieme12%40hotmail.com`
+    );
+    expect(res.status).toBe(403);
```
martijnvdbrug, line 467: The code above says `res.status(500).end();`, but the test says the response returns a 403 (forbidden). Is the testcase incorrect perhaps? Or is something else going on?

martijnvdbrug (changes requested):
Some questions and remarks. 

If you've fixed the testcase for localFileStorage, and are sure it works, can you duplicate the logic to the other storage strategies? (S3 and Google Cloud)

packages/vendure-plugin-invoices/test/e2e.spec.ts
LazyClicks: hey martijn, the 500 status was there as a fail safe in case the file ever disappears between the exists check and the stream reading, although now that i think about it, it's dead code at this point cause i don't see how it could be realistically triggered, imma remove it for better clarity on whats happening

packages/vendure-plugin-invoices/src/services/invoice.service.ts
LazyClicks: hey martijn, this is not needed as all storage strategies remove the temp file after a successful save
we only need to cover removal of temp file for error case because the removal is only triggered upon successful iteration

LazyClicks:
> Some questions and remarks.
> 
> If you've fixed the testcase for localFileStorage, and are sure it works, can you duplicate the logic to the other storage strategies? (S3 and Google Cloud)

both s3 and google cloud are remote storage strategies, so their download pdf hitting a missing invoice pdf wouldn't cause a server crash no?
i've added a simple try catch that uses safeRemove on the temp file only just in case

martijnvdbrug (approved):
Thanks for clarifying! Great work, approved and merged!

PR #726, opened 2026-08-11
feat: content health plugin

LazyClicks:
# Description

# To do before merge

# Breaking changes

# Screenshots

# Checklist

📌 Always:
- [ ] Set a clear title

👍 Most of the time:
- [ ] Updated the README

📦 For publishable packages:


martijnvdbrug (changes requested):
Great work! Some changes requested:

* Can you replace Axios with built in node fetch? We would like to minimize the use of external packages to reduce attack service. (For example, see this one from a while ago: https://www.trendaisecurity.com/en-us/resources-insights/trendai-security-blog/axios-npm-package-compromised) . We cant avoid external dependencies completely, but in this case we can easily replace it with node's fetch I believe.
* Set   `"private": false,` in package.json, so we can publish this package
* I don't think the custom fields `excludedFromContentChecks` should be part of the plugin: Consumers should exclude products/collections based on their own business rules. E.g. most clients have their own fields like `hidden` or `onlyDisplayAfterDate`. Introducing another customfield that has the same responsibility might be confusing. 

We should introduce a `shouldCheckEntity(product | collection)` that should dictate if an entity should be checked or not. This way consumers can exclude entities via their business rules

martijnvdbrug (approved):

PR #737, opened 2026-09-23
fix(utm-tracker): build dashboard assets

martijnvdbrug:
Ensures the UTM tracker package build includes its compiled Dashboard assets.

- Runs the existing Dashboard build as part of the package build
- Aligns the build flow with sibling plugins


martijnvdbrug:
No AI this isnt correct, you should have said dashboard build was already there!

